PR #15, opened 2016-07-08
De patching

stevector:


stevector:
Deprecated by https://github.com/stevector/search_api_pantheon/pull/19

PR #39, opened 2016-09-21
Switch to Terminus Alpha

stevector:
There are now two versions of Terminus, a 0.13 branch and an alpha release of an eventual 1.0. The 1.0 has a different command structure. This PR exists to verify that the 1.0 can do all of the tasks needed by this repo.


stevector:
Replaced by https://github.com/pantheon-systems/search_api_pantheon/pull/60

PR #53, opened 2016-10-24
Auto-post of schema

stevector:
https://www.drupal.org/node/2775549

Because this bug manifests on the creation of new multidevs, it is hard to test without significantly slowly down the test suite. Not sure what to do about that.

### DIFF
--- a/src/Plugin/SolrConnector/PantheonSolrConnector.php
+++ b/src/Plugin/SolrConnector/PantheonSolrConnector.php
@@ -141,7 +141,7 @@ class PantheonSolrConnector extends StandardSolrConnector {
     // When this plugin is reloaded, $this->configuration, will be repopulated
     // with $this->internalConfiguration().
     $this->configuration = $form_state->getValues();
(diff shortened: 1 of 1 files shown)


stevector:
Testing of this feature could be done by copying from one multidev to another: https://github.com/pantheon-systems/search_api_pantheon/pull/36/files

But again, I'm not sure if it is worth the additional time per build.

arosboro:
@stevector This looks like it would work in theory.  Are you using it on Pantheon with any projects successfully?

stevector:
@arosboro, I don't have any sites that need it / use it. Let me know if you can test the PR with a real site.

dalin-:
I've set this up on three separate D8 sites.  I'll report back if there's any adverse affects.

NickDickinsonWilde:
I can confirm that this fixes multidev sites to work without messing around. 👍

stevector:
Thanks @arosboro, @dalin, @NickWilde1990!

PR #73, opened 2019-02-11
Create LICENSE

davidstrauss:


stevector:
@davidstrauss this is a Drupal modules and Drupal modules have to be GPL, yes?

davidstrauss:
I think that's debatable, but I would go with GPLv2 for a Drupal module. I'll create a new PR.

PR #78, opened 2021-04-21
Upgrade for Search API Solr 4 (and Drupal 9)

dalin-:
I was just looking at this project the other day and noticed this phrase:

> Bug reports, feature requests, and feedback should be posted in the Drupal.org issue queue. For code changes, please submit pull requests against the GitHub repository rather than posting patches to drupal.org.

Dang!  A lot of work has already gone into porting this module for D9 over here:
https://www.drupal.org/project/search_api_pantheon/issues/3185736

So this pull request is just a copy of the code changes over there. 

Also, I'm copying over the outstanding issues:

> What's the original source of this `schema.xml` file? I'm guessing that it came from a schema file for Solr 4.x? That would explain why its version is `drupal-4.5-solr-4.x`. It seems to have some things that aren't compatible with Solr 3.x (`solr.SpatialRecursivePrefixTreeFieldType`). Pantheon's Solr doesn't complain, but Lando's Solr notices and fails silently. I modified this branch to remove those things (99.99% of Pantheon sites won't be trying to index geospatial polygons).

> Finally, it's important to truncate the solr index before you deploy this patch (and presumably an upgrade to D9). Otherwise you'll end up with 2x the number of documents in Solr, and regardless of what you do you can't get the old ones out. We'll need to document that somewhere. I guess in an UPGRADE.md file? Kinda easy to miss though.

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,7 @@ See the [Drupal.org for complete documentation on Search API](https://www.drupal
 
 ### Solr versions and schemas
 
-The version of Solr on Pantheon is Apache Solr v3.6. To accommodate this older version of Solr, use the 8.x-1.x branch of [Search API Solr](https://www.drupal.org/project/search_api_solr) and its Solr 4 schema file.
+The version of Solr on Pantheon is Apache Solr v3.6. To accommodate this older version of Solr, use the 4.x branch of [Search API Solr](https://www.drupal.org/project/search_api_solr) and its Search API Solr Legacy submodule and 3x schema file.
 
 ### Pantheon environments
 
--- a/composer.json
+++ b/composer.json
@@ -14,6 +14,6 @@
         }
     },
     "require": {
-        "drupal/search_api_solr": "~1.0-beta1"
+        "drupal/search_api_solr": "^4.0"
     }
 }
--- a/search_api_pantheon.info.yml
+++ b/search_api_pantheon.info.yml
@@ -1,7 +1,7 @@
 name: Search API Pantheon
 type: module
 description: Search API + Solr + Pantheon integration
-core: 8.x
+core_version_requirement: ^8 || ^9
 package: Pantheon
 dependencies:
-  - search_api_solr:search_api_solr
+  - search_api_solr:search_api_solr_legacy
--- a/search_api_pantheon.install
+++ b/search_api_pantheon.install
@@ -1,0 +1,14 @@
+<?php
+
+/**
+ * @file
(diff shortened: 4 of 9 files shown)


bdragon28:
I had independently rebased the same stuff myself privately for a site but hadn't gotten around to submitting it anywhere.  Here's the entire diff between my version (a) and this one (b):

```
diff --git a/src/Plugin/SolrConnector/PantheonSolrConnector.php b/src/Plugin/SolrConnector/PantheonSolrConnector.php
index 508dbe9..4e1f0ab 100644
--- a/src/Plugin/SolrConnector/PantheonSolrConnector.php
+++ b/src/Plugin/SolrConnector/PantheonSolrConnector.php
@@ -1,14 +1,8 @@
 <?php
 
-/**
- * @file
- * Provide a connection to Pantheon's Solr instance.
- */
-
 namespace Drupal\search_api_pantheon\Plugin\SolrConnector;
 
 use Drupal\search_api_pantheon\Solarium\PantheonCurl;
-use Drupal\Core\Annotation\Translation;
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\search_api_solr\Annotation\SolrConnector;
 use Drupal\search_api_solr\Solarium\EventDispatcher\Psr14Bridge;
@@ -80,9 +74,9 @@ class PantheonSolrConnector extends Solr36Connector {
    */
   public function defaultConfiguration() {
     return [
-        'schema' => '',
-        'solr_version' => '3.6.2',
-      ] + parent::defaultConfiguration();
+      'schema' => '',
+      'solr_version' => '3.6.2',
+    ] + parent::defaultConfiguration();
   }
 
   /**
@@ -120,7 +114,7 @@ class PantheonSolrConnector extends Solr36Connector {
       '#type' => 'radios',
       '#title' => $this->t('Schema file'),
       '#options' => $this->findSchemaFiles(),
-      '#description' => $this->t('Select a Solr schema file to be POSTed to Pantheon\'s Solr server'),
+      '#description' => $this->t("Select a Solr schema file to be POSTed to Pantheon's Solr server"),
       '#default_value' => $this->configuration['schema'],
     ];
 
@@ -181,4 +175,5 @@ class PantheonSolrConnector extends Solr36Connector {
     $this->pingServer();
     return parent::getDataFromHandler($handler, $reset);
   }
+
 }
```

I like your version better overall.

Big +1 to this, hopefully it gets in ASAP.

stovak:
We're adding support for solr8 for d8 and d9 very soon. See other PR's.

PR #95, opened 2021-08-27
CMS-224 - Refactor SEARCH_API_PANTHEON 8.x to reduce redundant code

stovak:


src/Services/PantheonGuzzle.php
```diff
@@ -153,14 +153,21 @@ public function getPsr18Adapter(): AdapterInterface {
...
-      return $r->withUri($uri);
+    $exploded = explode('/', $path);
+    if (!in_array(trim($this->endpoint->getCore(), '/'), $exploded)) {
+       array_unshift($exploded, trim($this->endpoint->getCore(), '/'));
     }
-    return $r;
+    if (!in_array(trim($this->endpoint->getPath(), '/'), $exploded)) {
+      array_unshift($exploded, trim($this->endpoint->getPath(), '/'));
+    }
+    $exploded = array_filter($exploded, function($item){
```
greg-1-anderson: I think the behavior of this is the same as calling `array_filter` with a single parameter.

src/Services/PantheonGuzzle.php
```diff
@@ -153,14 +153,21 @@ public function getPsr18Adapter(): AdapterInterface {
...
   public function requestUriAlterForPantheonEnvironment(RequestInterface $r) {
+    $toAdd = '';
     $uri = $r->getUri();
     $path = $uri->getPath();
-    $endpointHasPath = strpos($path, $this->endpoint->getPath());
-    if (strpos($path, $this->endpoint->getPath()) === FALSE) {
-      $uri = $uri->withPath($this->endpoint->getPath() . $this->endpoint->getCore() . $path);
-      return $r->withUri($uri);
+    $exploded = explode('/', $path);
+    if (!in_array(trim($this->endpoint->getCore(), '/'), $exploded)) {
```
greg-1-anderson: Is `getCore()` guaranteed to never contain a `/`? Could you change the code to continue to use strpos on getCore and getPath, or was there some reason you switched to looking at the elements of the paths individually with in_array?

src/Plugin/SolrConnector/PantheonSolrConnector.php
```diff
@@ -53,300 +47,141 @@ public function __construct(
...
-   * @param \Symfony\Component\DependencyInjection\ContainerInterface $container
-   * @param array $configuration
-   * @param string $plugin_id
-   * @param mixed $plugin_definition
-   *
-   * @return \Drupal\search_api\Plugin\ConfigurablePluginBase|\Drupal\search_api_pantheon\Plugin\SolrConnector\PantheonSolrConnector|static
-   * @throws \Exception
+   * Prepares the connection to the Solr server.
    */
-  public static function create(
```
greg-1-anderson, line 68: Was there a reason you deleted the static factory method?

greg-1-anderson:
It looks like you converted from the Drupal coding standards to PSR-2, and now the style checks are failing. It seems that this repo should continue to use Drupal coding standards, since it is going to be published on drupal.org in addition to here.

If you do keep the coding style on PSR-2, you should update your style checks so that it is testing the right set of standards.

src/Services/PantheonGuzzle.php
stovak: it trims first and last. The main thing is the solarium library inconsistently has some path/handler values that start and/or end with "/" and some that do not. Trying to establish that all path/handler values that are sent to another function do not have a start and ending "/" so that they can be added anywhere without getting a double "//"

stovak:
@namespacebrian @jabxjab Can you guys take a look at this before I merge it. Next phase for this project is creating tests but it's working right now so I want to merge to fix an error in "EA1" and tag with EA2.

RoboFile.php
```diff
@@ -0,0 +1,18 @@
...
+ */
+class RoboFile extends \Robo\Tasks
+{
+    public function testInSite()
+    {
+        $site_under_test = substr(\uniqid('sap-test-'), 0, 12);
+
+        $this->_exec("drush site:install --account-name=demo --site-name={$project} --locale=en --yes  {$profile}");
+        $this->_exec('drush pm-enable --yes  redis search_api search_api_solr search_api_pantheon search_api_pantheon_admin devel devel_generate');
+
```
jabxjab: Unnecessary extra lines

composer.json
```diff
@@ -41,26 +41,27 @@
...
     },
+
     "scripts": {
         "code:test" : [
             "vendor/bin/phpunit -c ./phpunit.xml"
         ],
         "code:fix": [
             "vendor/bin/phpcs --config-set installed_paths vendor/drupal/coder/coder_sniffer",
-            "vendor/bin/phpcbf --standard=Drupal,DrupalPractice src/* tests/Unit/* modules/*"
+            "vendor/bin/phpcbf --standard=PSR12 src/* tests/Unit/* modules/*"
```
jabxjab: A Drupal module should conform to the `Drupal` coding standards

composer.json
```diff
@@ -41,26 +41,27 @@
...
         ],
         "code:lint": [
             "vendor/bin/phpcs --config-set installed_paths vendor/drupal/coder/coder_sniffer",
-            "vendor/bin/phpcs --standard=Drupal,DrupalPractice . --ignore=vendor"
+            "vendor/bin/phpcs --standard=PSR12 --ignore=vendor src/* tests/Unit/* modules/*"
```
jabxjab: A Drupal module should conform to the `Drupal` coding standards

modules/search_api_pantheon_admin/js/settings_browser.js
```diff
@@ -1,11 +1,11 @@
...
-  function ($, Drupal, window, document) {
-    console.log("javascript library loaded");
-    Drupal.behaviors.searchApiPantheonSettingsBrowser = {
-      attach: function (context, settings) {
-        console.log("javascript behavior attached");
-        console.log(drupalSettings.searchApiPantheonAdmin);
-      }
-    };
-  } (jQuery, Drupal, this, this.document)
+    function ($, Drupal, window, document) {
```
jabxjab: The whole (unnecessary) `settings_browser.js` needs to be removed

modules/search_api_pantheon_admin/src/Access/AdminAccessCheck.php
```diff
@@ -11,7 +11,8 @@
 /**
  * Provides an access check for the "Solr Admin" routes.
  */
-class AdminAccessCheck implements AccessInterface {
+class AdminAccessCheck implements AccessInterface
```
jabxjab: Here and (down below in many places) - Drupal coding standards violation

.editorconfig
```diff
@@ -8,7 +8,7 @@ root = true
 [*]
 end_of_line = LF
 indent_style = space
-indent_size = 2
+indent_size = 4
```
jabxjab: Drupal coding standards requires indent size of `2`

composer.json
```diff
@@ -41,26 +41,27 @@
...
+        "kint-php/kint": "@stable"
     },
     "require-dev": {
         "drupal/coder": "@stable",
         "psy/psysh": "@stable",
-        "kint-php/kint": "@stable",
         "phpunit/phpunit": "@stable",
         "squizlabs/php_codesniffer": "@stable"
     },
+
```
jabxjab: Extra line

modules/search_api_pantheon_admin/src/Form/PantheonSolrAdminForm.php
```diff
@@ -14,108 +14,117 @@
...
+     *   Form control array.
+     */
+    protected function getViewSolrFile(string $filename, string $contents, bool $open = false): array
+    {
+        $toReturn = [];
+
+        $toReturn[$filename] = [
+            '#type' => 'details',
+            '#title' => $filename,
+            '#open' => !is_array($open),
```
jabxjab: There is no possibility for `$open` to be an array (PHP fatal error in that case). Look like should be `'#open'  => $open` instead.

modules/search_api_pantheon_admin/src/Form/PantheonSolrAdminForm.php
```diff
@@ -14,108 +14,117 @@
...
+                    'file' => $filename,
+                ],
+            ]);
+
+            $display_name = ucwords($filename);
+            $form[$filename] = [
+                '#type' => 'details',
+                '#title' => $display_name,
+                '#group' => 'status',
+                '#weight' => substr($filename, 0, -3) === "xml" ? -10 : 10,
```
jabxjab: It is recommended to use single quotes where possible

modules/search_api_pantheon_admin/src/Form/PantheonSolrAdminForm.php
```diff
@@ -14,108 +14,117 @@
...
+            ];
+            $form[$filename] = array_merge(
```
jabxjab: Seems that too many `$filename` usages/repeats. The following should work:
```
$form[$filename][] = $this->getViewSolrFile($filename, $file_contents, $is_open);
```
and in `getViewSolrFile`
as `protected function getViewSolrFile(string $details_title, string $contents, bool $is_open = false): array`:
```
return [
  '#type' => 'details',
  '#title' => $details_title,
  '#open' => $is_open,
  [
    '#type' => 'markup',
     '#markup' => sprintf('<pre>%s</pre>', Html::escape($contents)),
  ],
];
```

modules/search_api_pantheon_admin/src/Form/PostSolrSchema.php
```diff
@@ -13,68 +13,74 @@
...
-   */
-  public function buildForm(array $form, FormStateInterface $form_state, ServerInterface $search_api_server = NULL) {
-    $messages = $this->schemaPoster->postSchema($search_api_server->id());
-    $form['results'] = [
-      '#markup' => implode('<br>', $messages),
-    ];
+    /**
+     * {@inheritdoc}
+     *
+     * @param array $form
```
jabxjab: No need to repeat `@params` and `@return` because of `{@inheritdoc}`

modules/search_api_pantheon_admin/src/Form/PostSolrSchema.php
```diff
@@ -13,68 +13,74 @@
...
+     *   The form structure.
+     * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
+     * @throws \Drupal\search_api\SearchApiException
+     * @throws \Drupal\search_api_solr\SearchApiSolrException
+     *
+     * @throws \Drupal\Component\Plugin\Exception\InvalidPluginDefinitionException
+     */
+    public function buildForm(array $form, FormStateInterface $form_state, ServerInterface $search_api_server = null)
+    {
+        $messages = $this->schemaPoster->postSchema($search_api_server->id());
```
jabxjab: A preferred way to render a list would be as follows:
```
$form['results'] = [
  '#theme' => 'item_list',
  '#list_type' => 'ul',
  '#title' => t('The results'),
  '#items' => $this->schemaPoster->postSchema($search_api_server->id()),
];       

return $form;
```

src/Plugin/SolrConnector/PantheonSolrConnector.php
```diff
@@ -34,675 +27,365 @@ class PantheonSolrConnector extends SolrConnectorPluginBase implements
...
+    /**
+     * Returns the default endpoint name.
+     *
+     * @return string
+     *   The endpoint name.
+     */
+    public function getDefaultEndpoint()
+    {
+        // @codingStandardsIgnoreLine
+        return \Drupal\search_api_pantheon\Services\Endpoint::$DEFAULT_NAME;
```
jabxjab:
```
...
use \Drupal\search_api_pantheon\Services\Endpoint as PantheonSearchApiEndpoint;
...
return PantheonSearchApiEndpoint::$DEFAULT_NAME
```

src/Plugin/SolrConnector/PantheonSolrConnector.php
```diff
@@ -34,675 +27,365 @@ class PantheonSolrConnector extends SolrConnectorPluginBase implements
...
+
+        $summary['@autocommit_time_seconds'] = $max_time / 1000;
+        $summary['@autocommit_time'] = $this->container
+            ->get('date.formatter')
+            ->formatInterval($max_time / 1000);
+        $summary['@deletes_total'] =
+            (
+                intval($summary['@deletes_by_id'])
+                + intval($summary['@deletes_by_query'])
+            ) ?? -1;
```
jabxjab: `intval($summary['@deletes_by_id']) + intval($summary['@deletes_by_query'])` can't resolve to `null`, so possibly that conversion into ternary operator is needed. Applies also to other `(int) <something> ?? -` appeared above.

src/Plugin/SolrConnector/PantheonSolrConnector.php
```diff
@@ -34,675 +27,365 @@ class PantheonSolrConnector extends SolrConnectorPluginBase implements
...
-        $this->container->get('messenger')->error($error);
-      }
+        if (empty($stats) || empty($indexStats)) {
+            return $summary;
+        }
+
+        $max_time = -1;
+        $update_handler_stats = $stats['UPDATE']['updateHandler']['stats'] ?? -1;
+        $summary['@pending_docs'] =
+            (int)$update_handler_stats['UPDATE.updateHandler.docsPending'] ?? -1;
```
jabxjab: Please see the comment below.

src/Plugin/SolrConnector/PantheonSolrConnector.php
```diff
@@ -34,675 +27,365 @@ class PantheonSolrConnector extends SolrConnectorPluginBase implements
...
+        if (
+            isset(
+                $update_handler_stats['UPDATE.updateHandler.softAutoCommitMaxTime']
+            )
+        ) {
+            $max_time =
+                (int)$update_handler_stats['UPDATE.updateHandler.softAutoCommitMaxTime'];
+        }
+        $summary['@deletes_by_id'] =
+            (int)$update_handler_stats['UPDATE.updateHandler.deletesById'] ?? -1;
```
jabxjab: Please see the comment below.

src/Plugin/SolrConnector/PantheonSolrConnector.php
```diff
@@ -34,675 +27,365 @@ class PantheonSolrConnector extends SolrConnectorPluginBase implements
...
+        $summary['@deletes_by_query'] =
+            (int)$update_handler_stats['UPDATE.updateHandler.deletesByQuery'] ?? -1;
```
jabxjab: Please see the comment below.

src/Plugin/SolrConnector/PantheonSolrConnector.php
```diff
@@ -34,675 +27,365 @@ class PantheonSolrConnector extends SolrConnectorPluginBase implements
...
-      'info' => $this->getSchemaVersion(TRUE),
-    ];
-
-    $core_info = $this->getCoreInfo(TRUE);
-    foreach ($core_info['core'] as $key => $value) {
-      if (is_string($value)) {
         $view_settings[] = [
-          'label' => ucwords($key),
-          'info' => $value,
+            'label' => 'Pantheon Sitename',
```
jabxjab: Use `t()` translate function

src/Plugin/SolrConnector/PantheonSolrConnector.php
```diff
@@ -34,675 +27,365 @@ class PantheonSolrConnector extends SolrConnectorPluginBase implements
...
+            'info' => $this->getEndpoint()->getCore(),
+        ];
+        $view_settings[] = [
+            'label' => 'Pantheon Environment',
```
jabxjab: Use `t()` translate function

src/Plugin/SolrConnector/PantheonSolrConnector.php
```diff
@@ -34,675 +27,365 @@ class PantheonSolrConnector extends SolrConnectorPluginBase implements
...
+            'info' => getenv('PANTHEON_ENVIRONMENT'),
+        ];
+        $view_settings[] = [
+            'label' => 'Schema Version',
```
jabxjab: Use `t()` translate function

src/Plugin/SolrConnector/PantheonSolrConnector.php
```diff
@@ -34,675 +27,365 @@ class PantheonSolrConnector extends SolrConnectorPluginBase implements
...
+                $current = $mbeans->current();
+                $mbeans->next();
+                if ($mbeans->valid() && is_string($current)) {
+                    $stats[$current] = $mbeans->current();
+                }
+            }
+            $indexResponse = $this->getStatsQuery('admin/luke') ?? ['index' => []];
+            $indexStats = $indexResponse['index'] ?? [];
+        } catch (\Exception $e) {
+            $this->container->get('messenger')->addError('Unable to get stats from server!');
```
jabxjab: Use `t()` translate function

src/Plugin/SolrConnector/PantheonSolrConnector.php
```diff
@@ -34,675 +27,365 @@ class PantheonSolrConnector extends SolrConnectorPluginBase implements
...
+     * @return array
+     *   Form render array.
+     */
+    public function buildConfigurationForm(
+        array $form,
+        FormStateInterface $form_state
+    ) {
+        $form['notice'] = [
+            '#markup' =>
+                "<h3>All options are configured using environment variables on Pantheon.io's custom platform</h3>",
```
jabxjab: Use `t()` translate function

src/Plugin/SolrConnector/PantheonSolrConnector.php
```diff
@@ -34,675 +27,365 @@ class PantheonSolrConnector extends SolrConnectorPluginBase implements
...
+    /**
+     * Reaload the Solr Core.
+     *
+     * @return bool
+     *   Success or Failure.
+     */
+    public function reloadCore()
+    {
+        $this->logger->notice(
+            'Reload Core action for Pantheon Solr is automatic when Schema is updated.'
```
jabxjab: Use `t()` translate function

src/Services/PantheonGuzzle.php
```diff
@@ -26,141 +26,156 @@
...
+         * ), 'logger'
+         * );
+         **/
+        $cert = $_SERVER['HOME'] . '/certs/binding.pem';
+        $config = [
+            'base_uri' => $endpoint->getBaseUri(),
+            'http_errors' => false,
+            // Putting `?debug=true` at the end of any Solr url will show you the low-level debugging from guzzle.
+            // @codingStandardsIgnoreLine
+            'debug' => (PHP_SAPI == 'cli' || isset($_GET['debug'])),
```
jabxjab: `'debug' => php_sapi_name() === 'cli' || isset($_GET['debug']),`

src/Services/PantheonGuzzle.php
```diff
@@ -26,141 +26,156 @@
...
-  }
 
+    /**
+     * Request Middleware Callback.
+     *
+     * @param \Psr\Http\Message\RequestInterface $r
+     *
+     * @return \Psr\Http\Message\RequestInterface
+     */
+    public function requestUriAlterForPantheonEnvironment(RequestInterface $r)
```
jabxjab: I would recommend renaming `$r` into meaningful `$request`

src/Services/PantheonGuzzle.php
```diff
@@ -26,141 +26,156 @@
...
+    {
+        $toAdd = '';
+        $uri = $r->getUri();
+        $path = $uri->getPath();
+        $exploded = explode('/', $path);
```
jabxjab: ~$exploded~ $path_parts

src/Services/PantheonGuzzle.php
```diff
@@ -26,141 +26,156 @@
...
+        $shouldBeInUrl = $this->endpoint->getMySitename();
+        $shouldBeInPath = $this->endpoint->getPath();
+        if (!in_array(trim($shouldBeInUrl, '/'), $exploded)) {
+            array_unshift($exploded, trim($this->endpoint->getCore(), '/'));
+        }
+        if (!in_array(trim($shouldBeInPath, '/'), $exploded)) {
+            array_unshift($exploded, trim($this->endpoint->getPath(), '/'));
+        }
+        $exploded = array_filter($exploded, function ($item) {
+            return !empty($item);
```
jabxjab:
```
$path_parts = array_filter($path_parts);
```

src/Services/SchemaPoster.php
```diff
@@ -23,369 +23,382 @@
...
-      ]);
-      $this->logger->error($message);
+            ? 'notice'
+            : 'error';
+        $this->logger->{$log_function}('Files uploaded: {status_code} {reason}', [
+            'status_code' => $response->getStatusCode(),
+            'reason' => $response->getReasonPhrase(),
+        ]);
+        $message = vsprintf('Result: %s Status code: %d - %s', [
+            $log_function == 'error' ? 'NOT UPLOADED' : 'UPLOADED',
```
jabxjab: `t()`

src/Services/SchemaPoster.php
```diff
@@ -23,369 +23,382 @@
...
-      return $response->getBody();
-    }
-    catch (\Throwable $e) {
-      $message = vsprintf('File: %s, Status code: %d - %s', [
-        'filename' => $filename,
-        'status_code' => $e->getCode(),
-        'reason' => $e->getMessage(),
-      ]);
-      $this->logger->error($message);
+            ? 'notice'
```
jabxjab: Better to use `info` instead of `notice` for OK requests

src/Services/SchemaPoster.php
```diff
@@ -23,369 +23,382 @@
...
+            $response = $this->uploadSchemaFiles($this->getSolrFiles($server_id));
+        }
+        // LOCAL DOCKER.
+        if (isset($_SERVER['ENV']) && $_SERVER['ENV'] === 'local') {
+            $response = $this->uploadSchemaAsZip($server_id);
+        }
+        if (!$response instanceof Response) {
+            throw new \Exception('Cannot post schema to environment url.');
+        }
+        $log_function = in_array($response->getStatusCode(), [
```
jabxjab: Useless extra check, `guzzle` will throw an exception for anything non-2xx. Should be converted int `try/catch` block.

src/Services/SchemaPoster.php
```diff
@@ -23,369 +23,382 @@
...
+            json_encode($toSend)
+        );
+        $response = $this->getClient()->sendRequest($request);
 
-    if (!$server instanceof EntityInterface) {
-      throw new \Exception(
-        'cannot retrieve the solr server connection settings from the database'
-      );
+        // Parse the response.
+        $log_function = in_array($response->getStatusCode(), [200, 201, 202, 203])
```
jabxjab: Useless extra check, `guzzle` will throw an exception for anything non-2xx. Should be converted int `try/catch` block.

src/Services/SchemaPoster.php
```diff
@@ -23,369 +23,382 @@
...
+                        'overwrite' => 'true',
+                    ],
+                    'headers' => [
+                        'Content-Type' => 'application/octet-stream',
+                    ],
+                    'body' => $contents,
+                ]
+            );
+            // Parse the response.
+            $log_function = in_array($response->getStatusCode(), [200, 201, 202, 203])
```
jabxjab: Useless extra check, `guzzle` will throw an exception for anything non-2xx. Should be converted int `try/catch` block.

jabxjab (changes requested):

RoboFile.php
stovak: removed it. You're right we should do testing in a completely different ticket.

modules/search_api_pantheon_admin/js/settings_browser.js
stovak: removed.

src/Services/PantheonGuzzle.php
greg-1-anderson: Yes, but the old code works with paths like `foo/bar`, whereas the new code would fail. If `getCore()` never contains an *interior* `/`, then that's fine; in that instance, the new code is clearer. The old code is safer if there might ever in the future be an interior `/` in the `getCore()` path.

greg-1-anderson (changes requested):
I agree with @jabxjab's review comments on Drupal Coding conventions et. al.; otherwise looking good to me.

greg-1-anderson:
If you can't get the code beautifier to easily bring you back to Drupal Coding standards, we could defer this for follow-on work.

src/Services/PantheonGuzzle.php
stovak: Changed it to search for site_id in the exploded array.

src/Services/SchemaPoster.php
stovak: No, it actually doesn't. if you look at the config HTTP_ERRORS => false will keep It from throwing exceptions for non-200 responses.

src/Services/SchemaPoster.php
stovak: See above.

src/Services/SchemaPoster.php
jabxjab: > No, it actually doesn't. if you look at the config HTTP_ERRORS => false will keep It from throwing exceptions for non-200 responses.

What is the point in resetting this default config? It is safer to follow the default one (and use `try/catch`)

.phpcs.xml
```diff
@@ -0,0 +1,363 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jabxjab, line 1: What is the purpose of this file? 
A proper way to run `phpcs` is `vendor/bin/phpcs --standard=Drupal .`

jabxjab (changes requested):

.phpcs.xml
stovak: It's an abbreviated Drupal coding standard that I copied from core. It allows you to forgo some of the redundant Documentation of variables. One of my pet peeves about the drupal standard is it requires all this redundant documentation of param and option variables that doesn't really add anything to the understanding of the code. If a param is a class type, it really shouldn't need a comment. The type of PantheonGuzzle is sufficient for the reader of the code to understand what the variable is and what it's used for.

jabxjab: Changing the existing Drupal standards check list makes it non-Drupal compliant anymore.

PR #97, opened 2021-09-08
Starting on End-to-End Tests

stevector:
I'm starting on this PR as a means of validating our internal docs on how to get search functioning on an environment. I think the error currently happening in CI when enabling the relevant modules is Solr8 might not actually be enabled on the environment I'm testing: https://pantheon.slack.com/archives/C02455RD3EG/p1631138245107200?thread_ts=1631037981.086600&cid=C02455RD3EG

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -19,7 +19,9 @@ jobs:
       TERMINUS_TOKEN: ${{ secrets.TERMINUS_TOKEN }}
       TERMINUS_TOKEN_2: ${{ secrets.TERMINUS_TOKEN_2 }}
       GITHUB_RUN_NUMBER: ${{ github.run_number }}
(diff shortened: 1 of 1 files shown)


stovak:
FYI: I think what we want to do is to use a custom distro for testing that has search_api_pantheon built into it and is only available in the CI org. 

I also think we want to use ROBO to automate the dirty work.

stevector:
👍 I think it's up to you and those working on automated test across the board to make that call. Out of curiosity, what benefit will be had from abstracting away the bash commands into ROBO? I think the main benefit of straight bash is that the automated tests can function as an implementation of set up documentation that our customers would use. Do you expect "use this ROBO command" would become the thing we'd direct customers to do?

stovak:
Automated tests working with Another PR.

PR #100, opened 2021-09-20
Add a Hook_requirements that prevents search_api_pantheon from being enabled unless env vars are set

stovak:


RoboFile.php
```diff
@@ -0,0 +1,372 @@
...
+    return NULL;
+  }
+
+  /**
+   * @param string $site_name
+   * @param string $env
+   */
+  public function testEnvSolr(string $site_name, string $env = 'dev') {
+    $site_folder = $this->getSiteFolder($site_name);
+    $pantheon_yml_contents = Yaml::parseFile($site_folder . '/pantheon.yml');
```
namespacebrian, line 365: errored out here..
```
PHP Fatal error:  Uncaught Error: Class "Yaml" not found in /Users/brianweaver/repos/search_api_pantheon/RoboFile.php:365
```

should there be a `use` at the top of the file to include this class?

namespacebrian (approved):

PR #101, opened 2021-10-01
Add missing schema

jeffm2001:
Our test suite complains about a missing schema for `search_api.server.pantheon_solr8:backend_config.connector_config`. This takes care of it. 

The schema here mostly mirrors `search_api_solr.connector.standard.schema.yml`, with a couple of tweaks to match what's actually in `search_api.server.pantheon_solr8.yml`


heddn:
I think this looks good. RTBM.

heddn:
Adding a functional test to the module and a test module that installs and configures an index would be a great way to confirm the schema issues do not return.

stovak:
rerolled https://github.com/pantheon-systems/search_api_pantheon/pull/165/files

PR #107, opened 2021-10-21
[CMS-170] Add automated testing

jazzsequence:
Finishes the Robo workflow tests to verify Solr8 is installed and working as expected.
Those workflows are then used as GitHub actions to run the same tests on a non-local environment.


RoboFile.php
```diff
@@ -30,6 +31,7 @@ class RoboFile extends Tasks {
    */
   public function __construct() {
     $this->started = new DateTime();
+    require_once 'vendor/autoload.php';
```
jazzsequence, line 34: This seems like it shouldn't be required, but the GitPHP library doesn't seem to be autoloaded otherwise without explicitly including the `autoload.php`.

RoboFile.php
```diff
@@ -102,7 +102,7 @@ public function testCreateSite(string $site_name, array $options = ['org' => NUL
     if (empty($site_info)) {
       $toReturn = $this->taskExec(static::$TERMINUS_EXE)
         ->args('site:create', $site_name, $site_name, 'drupal9');
-      if ($options['org'] !== NULL) {
+      if ( !empty( $options['org'] ) ) {
```
jazzsequence, line 111: Previous iteration returned an undefined array offset error because `$options['org']` was assigned `null`.

RoboFile.php
```diff
@@ -47,11 +49,9 @@ public function testFull(string $site_name = NULL) {
     $this->testCloneSite($site_name);
     $this->testRequireSolr($site_name, 'dev-' . $branch);
     $this->testGitPush($site_name);
-    $this->waitForWorkflow($site_name);
     $this->testConnectionGit($site_name, 'dev', 'sftp');
     $this->testSiteInstall($site_name);
     $this->testConnectionGit($site_name, 'dev', 'git');
-    $this->waitForWorkflow($site_name);
     try {
       // This should fail:
```
jazzsequence: Todo: this comment can specify that it will fail because solr isn't enabled for the site yet, and solr 8 isn't enabled

RoboFile.php
```diff
@@ -63,7 +63,7 @@ public function testFull(string $site_name = NULL) {
     $this->setSiteSearch($site_name, 'enable');
     $this->testEnvSolr($site_name);
     $this->testGitPush($site_name);
-    $this->waitForWorkflow($site_name);
+    $this->testConnectionGit($site_name, 'dev', 'sftp');
```
jazzsequence: This can go away
```suggestion
```

stovak (approved):

PR #113, opened 2021-11-04
CMS-387: Add new select command.

kporras07:
Add new `saps` command.


stovak:
I rebased around Chris' new automated testing. Let's see if it passes....

stovak (approved):

PR #114, opened 2021-11-04
Switch docker image to new 8.x-php8.4 tag.

kporras07:
I think we should be using a Pantheon official image and ideally, that image should contain terminus and build tools in it. I'm switching the image in this PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ jobs:
   build_test:
     runs-on: ubuntu-latest
     container:
-      image: quay.io/pantheon-public/build-tools-ci:7.4.x
+      image: quay.io/pantheon-public/build-tools-ci:8.x-php7.4
       options: --user root
     name: Build and test
     env:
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -41,23 +41,10 @@ jobs:
       - name: Code sniff
         run: composer run-script code:lint
 
-      - name: Install Terminus
```
jazzsequence, line 44: Is terminus already installed at another step?

.github/workflows/ci.yml
```diff
@@ -41,23 +41,10 @@ jobs:
...
-        run: |
-          wget https://github.com/pantheon-systems/terminus/releases/download/3.0.0-beta1/terminus
-          chmod +x terminus
-          mv terminus /usr/local/bin/t3
-          mkdir ~/.terminus
-          t3 self:info
-
-      - name: Install Terminus Build Tools
-        run: |
-          t3 self:plugin:install terminus-build-tools-plugin
```
jazzsequence, line 54: These are used so in some of the Robo workflows. Is there some other place that the build tools are being installed?

.github/workflows/ci.yml
kporras07: It's in the docker image

.github/workflows/ci.yml
kporras07: It's in the docker image

jazzsequence (approved):
If these things are already in the image then this looks g2g to me. 👍

PR #115, opened 2021-11-04
CMS-395: Select test.

kporras07:
- Added workaround for issue with base image not having terminus plugins installed.
- Added collision protection for parallel runs (I found once that 2 different runs were acting over the same site)
-  Updated getCurrentBranch to getCurrentConstraint to handle different branches/PRs scenarios to always get the right version constraint
- Added select command test (and added a Solr index before it)
- Make select command return instead of logging with notice level


RoboFile.php
```diff
@@ -70,15 +74,42 @@ public function testFull(string $site_name = NULL) {
...
     $this->output()->write( 'All done! 🎉' );
     return ResultData::EXITCODE_OK;
   }
 
   /**
    *
    */
-  protected function getCurrentBranch(): ?string {
-    return trim(shell_exec('git rev-parse --abbrev-ref HEAD'));
+  protected function getCurrentConstraint(): string {
```
jazzsequence, line 93: Nit: It would be good if we added some function docs rather than empty comment blocks.

jazzsequence (approved):
Lgtm! It would be good to have more inline docs but that's not a blocker. Ship it!

RoboFile.php
kporras07: I'm going ahead and merging this but definitely will add more docs in my next PR. Thanks!

PR #116, opened 2021-11-05
Remove unneeded elements from PantheonEventDispatcher class.

kporras07:
I found a very consistent issue when a new site gets created with the latest 8.x version of this module.

In this situation, when you index Solr, it never gets committed.

I was able to track this behavior down to this class and then I found that removing this function fixes the issue.

Was there any reason this function was implemented like this? Anything I should be aware of?


stovak:
This was a back port I added during the D8 compatibility effort. As long as the module works with d8 and D9, make whatever changes are necessary to complete the task.

kporras07:
@stovak this is already tested in both D8 and D9 and it's now working as expected. Please take a look

stovak:
Great work. Very well done.

PR #121, opened 2021-11-11
Remove dependencies that are already added by core.

damienmckenna:
There's no reason to include dependencies in composer.json for things that core already adds.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -37,10 +37,8 @@
     "require": {
         "php": ">=7.4",
         "ext-curl": "*",
-        "ext-json": "*",
         "ext-zip": "*",
         "drupal/search_api_solr": "^4.2.1",
-        "guzzlehttp/guzzle": "^6",
         "http-interop/http-factory-guzzle": "^1.0",
         "php-http/guzzle6-adapter": "^2.0",
         "psr/event-dispatcher": "^1.0"


composer.json
```diff
@@ -37,10 +37,8 @@
     "require": {
         "php": ">=7.4",
         "ext-curl": "*",
-        "ext-json": "*",
         "ext-zip": "*",
```
stovak, line 40: What if one day they're not provided by core? They are explicit dependencies of this module so, IMHO, they should be declared in the composer file. @greg-1-anderson @kporras07 @namespacebrian how do you guys feel about this?

damienmckenna: FYI Drupal 10 is going to require JSON storage in the database, so the requirement on JSON is going to get more important, not less.

greg-1-anderson: In theory, Drupal core cannot drop a requirement or bump up to a higher major version of guzzle without also bumping its major version (that is, one should expect that the major version of guzzle should not change until D10). So, in theory, it's fine to remove these constraints here.

However, if in practice Drupal did make such a change, the explicit dependency on guzzle here would prevent the use of this module with an incompatible version of guzzle. Since the inclusion of the explicit dependency here is not causing any harm, as far as I can tell, my inclination is to say we're probably better off leaving them in place, as documentation if for no other reason.

stovak: Ok. I think I'm voting to leave both of these in the composer file, so going to close this without merging.

PR #129, opened 2021-11-24
CMS-426: Processors documentation

stovak:


src/Plugin/SolrConnector/PantheonSolrConnector.php
```diff
@@ -404,6 +405,7 @@ protected function getStatsQuery(string $handler) {
                               'accept' => 'application/json',
                               'contenttype' => 'application/json',
                               'json.nl' => 'flat',
+                              'commit' => 'true',
```
kporras07: Solarium usually adds commit_within parameter to the requests. Is there a reason this needs commit? I feel like this may be hiding a different problem. Besides, this looks like a GET request that is forcing committing, is that expected?

stovak: So the reason I added this is because there are some stats queries that were failing because there were outstanding changes that needed to be committed. It fixes the issue where you see there are "0" documents and you've just sent all your data to the server.

kporras07: Hmmm, to me that issue is always fixed with page reload. Actually, I don't consider that an issue of search_api_pantheon because it happens with every Solr server that I've used (local or from any provider)

stovak: Going to remove it until I have a better argument for it being there.

kporras07 (approved):
Looks good

PR #132, opened 2021-12-09
[CMS-447, CMS-317] Add `*_timeout` config fields, prevent env vars from being exported to config.yml

jabxjab:


src/Plugin/SolrConnector/PantheonSolrConnector.php
```diff
@@ -120,19 +120,43 @@ public static function create(
...
-          'scheme' => getenv('PANTHEON_INDEX_SCHEME'),
-          'host' => getenv('PANTHEON_INDEX_HOST'),
-          'port' => getenv('PANTHEON_INDEX_PORT'),
-          'path' => getenv('PANTHEON_INDEX_PATH'),
-          'core' => getenv('PANTHEON_INDEX_CORE'),
-          'schema' => getenv('PANTHEON_INDEX_SCHEMA'),
-          'solr_version' => '8',
-          'commit_within' => 1000,
-      ]);
+    return array_merge(
```
stovak, line 155: I like this approach. Great work!

src/Services/Endpoint.php
```diff
@@ -42,19 +44,31 @@ class Endpoint extends SolariumEndpoint {
...
-      'scheme' => getenv('PANTHEON_INDEX_SCHEME'),
-      'host' => getenv('PANTHEON_INDEX_HOST'),
-      'port' => getenv('PANTHEON_INDEX_PORT'),
-      'path' => getenv('PANTHEON_INDEX_PATH'),
-      'core' => getenv('PANTHEON_INDEX_CORE'),
-      'schema' => getenv('PANTHEON_INDEX_SCHEMA'),
-      'collection' => NULL,
-      'leader' => FALSE,
-    ]);
+    $storage = \Drupal::entityTypeManager()->getStorage('search_api_server');
```
kporras07: It would be better to receive entity_type.manager service as argument in this __construct.

src/Services/Endpoint.php
```diff
@@ -42,19 +44,31 @@ class Endpoint extends SolariumEndpoint {
...
+    /** @var \Drupal\search_api\ServerInterface $server */
+    $server = $storage->load(self::DEFAULT_NAME);
```
kporras07: I'm not sure what's the right approach, but I'd like to raise a concern about this line forcing users to create the server with a "hardcoded" machine name. This is not common practice in other Solr providers and could lead to lots of problems from customers.

What if they screw up the server that was created during the module installation and then decide to create a new one and the name is different?

What if they're updating from old search_api_pantheon and then the config is not installed and their server is named different?

src/Services/Endpoint.php
jabxjab: @kporras07 agree, added

src/Services/Endpoint.php
jabxjab: My understanding is that `search_api_pantheon` module must provide a single predefined server configuration for the platform, so it fits this hardcoded approach we currently have well. 
If this is the case, my main concern would be that we missing the following features:
* prevent a user from deleting the server config (ie. "delete" context action)
* prevent a user from creating an additional `Pantheon Search Connector`-based server config (ie. "Add Server")

@stovak what are your thoughts on this?

stovak:
@jabxjab failing tests with 
```
Error: Call to a member function getBackendConfig() on null in Drupal\search_api_pantheon\Services\Endpoint->__construct() (line 57 of /code/web/modules/composer/search_api_pantheon/src/Services/Endpoint.php) #0 /code/web/core/lib/Drupal/Component/DependencyInjection/Container.php(262): Drupal\search_api_pantheon\Services\Endpoint->__construct(Array, Object(Drupal\Core\Entity\EntityTypeManager))
```

jabxjab:
> @jabxjab failing tests with
> 
> ```
> Error: Call to a member function getBackendConfig() on null in Drupal\search_api_pantheon\Services\Endpoint->__construct() (line 57 of /code/web/modules/composer/search_api_pantheon/src/Services/Endpoint.php) #0 /code/web/core/lib/Drupal/Component/DependencyInjection/Container.php(262): Drupal\search_api_pantheon\Services\Endpoint->__construct(Array, Object(Drupal\Core\Entity\EntityTypeManager))
> ```

@stovak fixed

stovak (approved):

PR #136, opened 2022-02-11
Use ZK if available to update config files

stovak:


src/Services/SchemaPoster.php
```diff
@@ -78,18 +76,17 @@ public function __construct(
    * @throws \Drupal\search_api_solr\SearchApiSolrException
    */
   public function postSchema(string $server_id): array {
+    // LOCAL DOCKER.
+    if (extension_loaded('zookeeper') && isset($_SERVER['PANTHEON_CONFIGSERVER_HOST']) && isset($_SERVER['PANTHEON_CONFIGSERVER_PORT'])) {
```
greg-1-anderson, line 80: Should these be `PANTHEON_` environment variables? i.e., are they set by the platform? If not, they should probably have a different prefix.

src/Services/SchemaPoster.php
```diff
@@ -98,18 +95,49 @@ public function postSchema(string $server_id): array {
...
+   * @throws \ZookeeperException
+   * @throws \Drupal\search_api\SearchApiException
+   * @throws \Drupal\search_api_solr\SearchApiSolrException
+   * @throws \Drupal\Component\Plugin\Exception\PluginNotFoundException
+   */
+  protected function updateZookeeperConfigs(string $server_id) {
+    $message = '';
+    $schemaFiles = $this->getSolrFiles($server_id);
+    $projectName = getenv('PROJECT_NAME');
+    $zk = new \Zookeeper(getenv('PANTHEON_CONFIGSERVER_HOST') . ":" . getenv('PANTHEON_CONFIGSERVER_PORT'));
```
greg-1-anderson, line 123: Maybe pass in host and port as parameters, and read the env vars from the caller?

greg-1-anderson:
There isn't a lot of extra code here, and it's loosely coupled with the rest of the code, so it's certainly something we could consider including.

What's missing, though, is documentation that explains why you would want to use zookeeper, how you set up zookeeper, and how your config management is different when using zookeeper vs not.

Is zookeeper is only available via ext-zookeeper? Is there a way to use this to manage config on your live site, or is this only for local development? If this isn't for use on Pantheon, then it further confuses the use case; this also should be explained in the documentation.

PR #142, opened 2022-06-08
Force cleanup command

kporras07:
Adds new search-api-pantheon:force-cleanup that will delete everything in the given server no matter index_id or hash.

This is useful for situations like following:
1) User created an index, messed things up and deleted the index before deleting the content (that content will be stuck there forever)
2) search_api_solr.site_hash state changed (maybe because db was copied from one env to another, or for some other unknown reason)

Under these situations, going to the server and clicking the "Delete all server contents" button really doesn't work because that's filtered by hash and index_id. This new command does the same but removing those filters so it will effectively clean the server contents.


stovak (approved):
Great work!

PR #143, opened 2022-06-29
[CMS-728] Allow users to post custom schema files.

kporras07:


RoboFile.php
```diff
@@ -302,6 +303,34 @@ public function testCloneSite(string $site_name) {
...
+  /**
+   * Add allow plugins section to composer.
+   *
+   * @param string $site_name
+   *   The machine name of the site to add the allow plugins section to.
+   */
+  public function testAllowPlugins(string $site_name) {
+    $site_folder = $this->getSiteFolder($site_name);
+    chdir($site_folder);
+    $plugins = [
```
jabxjab, line 315: Not sure if this should match, but the list of plugins currently allowed by the platform (IC) is as follows:
* composer/installers
* cweagans/composer-patches
* drupal/core-composer-scaffold
* dealerdirect/phpcodesniffer-composer-installer
* [not applicable to Drupal] johnpbloch/wordpress-core-installer

modules/search_api_pantheon_admin/src/Form/PostSolrSchema.php
```diff
@@ -49,18 +57,83 @@ public function getFormId() {
...
+   */
+  protected function getMessageFunction(string $type) {
+    $functions = [
+      'info' => 'addStatus',
+      'error' => 'addError',
+    ];
+    if (isset($functions[$type])) {
+      return $functions[$type];
+    }
+    else {
```
jabxjab: `else` is redundant

RoboFile.php
greg-1-anderson: I think it only matters that this function converts the test fixture to a functioning site, so only plugins used in the fixtures need be listed.

greg-1-anderson (approved):

PR #146, opened 2022-11-16
Drupal 10 compatibility fixes

ossobuffo:
See comment here:
https://www.drupal.org/project/search_api_pantheon/issues/3289519#comment-14789310

I determined that the Guzzle6 adapter was not being used anymore, so I have bumped the Guzzle requirement to allow Drupal 10's Guzzle 7.5.0. I believe this should get us to full Drupal 10 compatibility. 

The Project Update Bot also changed a few lines in the unit tests, but I suspect that would cause those tests to fail against older Drupal, so I have reverted those changes.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -42,10 +42,9 @@
         "ext-json": "*",
         "ext-zip": "*",
         "drupal/search_api_solr": "^4.2",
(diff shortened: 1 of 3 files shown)


jspellman814:
Closing in favor of #147

PR #148, opened 2022-11-18
[CMS-1125] More D10 compatibility fixes.

kporras07:


src/Solarium/EventDispatcher/Psr14Bridge.php
```diff
@@ -20,7 +19,18 @@ public function __construct(ContainerAwareEventDispatcher $eventDispatcher) {
     $this->dispatcher = $eventDispatcher;
   }
 
-  public function dispatch($event, Event $null = NULL) {
+  /**
+   * Call magic method to account for Symfony\Contracts vs Symfony\Component method declarations.
+   */
+  public function __call($name, $args) {
```
kporras07, line 25: I don't super love this but it works. I'm open to suggestions :)

damienmckenna: We could just drop compatibility with the older version and bump the core requirement.

modules/search_api_pantheon_admin/search_api_pantheon_admin.info.yml
```diff
@@ -1,7 +1,7 @@
 name: Search API Pantheon Admin
 type: module
 description: Administer a Pantheon Search server
-core_version_requirement: ^8.8 || ^9
+core_version_requirement: ^8.8 || ^9 || ^10
```
damienmckenna, line 4: I think we should drop D8 support.

composer.json
```diff
@@ -46,7 +46,7 @@
         "http-interop/http-factory-guzzle": "^1.0",
         "kint-php/kint": "^4.1",
         "psr/event-dispatcher": "^1.0",
-        "symfony/finder": "^4|^5"
+        "symfony/finder": "^3|^4|^5|^6"
```
damienmckenna, line 49: D9.3 requires symfony/finder 4.4, so the ^3 bit should be removed.

src/Solarium/EventDispatcher/Psr14Bridge.php
damienmckenna: Per https://www.drupal.org/node/3159012 this change was introduced in 9.1.0.

damienmckenna (changes requested):
I think we should drop support for D8, and bump the 9.x requirement to 9.1 so that the Symfony class change can be ignored, see https://www.drupal.org/node/3159012 for details.

greg-1-anderson (changes requested):
I think that bumping the major version and dropping support for Drupal 8 as @damienmckenna suggests makes the most sense.

If we did decide to maintain Drupal 8 compatibility, I'd prefer that we did so by loading a different php file for Drupal 8 and Drupal 9, rather than rely on a magic `__call()` method. Since Drupal 8 is EOL, just dropping support is a reasonable alternative to that.

kporras07:
Closing in favor of #152

PR #147, opened 2022-11-16
Drupal 10 compatibility fixes

jspellman814:
This is a clone of #146 so our tests will run. See comment [here](https://github.com/pantheon-systems/search_api_pantheon/pull/146#issue-1452319105) for more details on these changes.

Props to @ossobuffo

PR #150, opened 2022-11-21
Revert PR #147.

kporras07:
Because adding D10 compatibility to this same branch has proven to be harder than expected. A new 9.x branch will be created shortly to replay these changes on top of it.


jspellman814 (approved):

damienmckenna:
There is zero reason to create a new 9.x branch, it will just cause more problems than it's worth, see https://medium.com/jakob-on-drupal/dont-go-making-major-version-changes-474293dda1d7 for a good writeup on the topic.

PR #163, opened 2023-02-14
[BUGS-5696] Separate index item to its own command.

kporras07:


stovak (approved):
LGTM when tests pass. This is a good decision to split this into two.

PR #165, opened 2023-03-29
Jeffm2001 8.x schema

stovak:
Fix for schema issues contrib by jeffm2001

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -55,7 +55,8 @@
         "drupal/coder": "@stable",
         "phpunit/phpunit": "@stable",
         "psy/psysh": "@stable",
-        "squizlabs/php_codesniffer": "@stable"
+        "squizlabs/php_codesniffer": "@stable",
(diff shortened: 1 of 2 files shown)


namespacebrian (approved):

composer.json
```diff
@@ -55,7 +55,8 @@
         "drupal/coder": "@stable",
         "phpunit/phpunit": "@stable",
         "psy/psysh": "@stable",
-        "squizlabs/php_codesniffer": "@stable"
+        "squizlabs/php_codesniffer": "@stable",
+        "symfony/yaml": "^4"
```
kporras07: We already require symfony/yaml in the "require" section. I don't think we need it here

stovak: Removed.

dpagini:
This fix is not actually working for me... I think this module should use the fix from:  https://www.drupal.org/files/issues/2022-09-12/3262003-7.patch

PR #172, opened 2023-08-23
LOPS-1666 - remove schema post from ping function

stovak:
Removing this function will use the function from the parent class.


kporras07 (approved):
Assuming you tested it enough, LGTM

PR #173, opened 2023-12-20
[BUGS-6560]: tag branches

rwagner00:
Per request in https://getpantheon.atlassian.net/browse/BUGS-6560, whenever a new tag (ex: 1.2.3) is pushed to the repo, if a branch 1.2.x does not yet exist, that branch will be created. If the branch does exist, then the branch will be updated with the most recent commits from the corresponding major version branch 1.x.

Action testing was done on a fork which can be found here: https://github.com/rwagner00/search_api_pantheon/actions
Testing was done there to avoid any potential for a workflow under development to deploy something it ought not.

### DIFF
--- a/.github/workflows/create-branch-on-tag.yml
+++ b/.github/workflows/create-branch-on-tag.yml
@@ -24,13 +24,14 @@ jobs:
           key: ${{ secrets.SSH_KEY }}
           known_hosts: ${{ secrets.KNOWN_HOSTS }}
           if_key_exists: ignore
-
       - name: Extract tag name
         id: tag_name
         run: |
           FULL_TAG=${GITHUB_REF#refs/tags/}
           MAJOR_MINOR="$(echo $FULL_TAG | cut -d '.' -f 1,2).x"
-          echo "::set-output name=BRANCH_NAME::$MAJOR_MINOR"
+          MAJOR_ONLY="$(echo $FULL_TAG | cut -d '.' -f 1).x"
+          echo "BRANCH_NAME=$MAJOR_MINOR" >> $GITHUB_OUTPUT
+          echo "MAJOR_BRANCH_NAME=$MAJOR_ONLY" >> $GITHUB_OUTPUT
       - name: Check if branch exists
         id: check_branch
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/create-branch-on-tag.yml
```diff
@@ -0,0 +1,63 @@
...
+          known_hosts: ${{ secrets.KNOWN_HOSTS }}
+          if_key_exists: ignore
+      - name: Extract tag name
+        id: tag_name
+        run: |
+          FULL_TAG=${GITHUB_REF#refs/tags/}
+          MAJOR_MINOR="$(echo $FULL_TAG | cut -d '.' -f 1,2).x"
+          MAJOR_ONLY="$(echo $FULL_TAG | cut -d '.' -f 1).x"
+          echo "BRANCH_NAME=$MAJOR_MINOR" >> $GITHUB_OUTPUT
+          echo "MAJOR_BRANCH_NAME=$MAJOR_ONLY" >> $GITHUB_OUTPUT
```
jazzsequence, line 34: These lines, as well as the ` >> $GITHUB_OUTPUT` lines on 42 and 45 are causing this warning:
`Warning: The `set-output` command is deprecated and will be disabled soon. Please upgrade to using Environment Files. For more information see: https://github.blog/changelog/2022-10-11-github-actions-deprecating-save-state-and-set-output-commands/`

I don't know what "soon" means given that the changelog linked is from 2022 but probably best to change the behavior anyway before the workflow breaks from under us.

jazzsequence (changes requested):

.github/workflows/create-branch-on-tag.yml
jazzsequence: Link: https://github.com/rwagner00/search_api_pantheon/actions/runs/7282073912/job/19843780177#step:3:13

rwagner00: That was fixed in the next commit, as you can see https://github.com/rwagner00/search_api_pantheon/actions/runs/7282092388 there are no more deprecation notices

jazzsequence (approved):

PR #176, opened 2024-03-05
CMSP-789: Config XML

rwagner00:
Resolves issue where search api was now sending bad data to the solrconfig.xml file.


stovak (approved):
This looks good, but make sure you test in a live solr environment.

BhavdeepSinghB (approved):

rwagner00:
Tested very thoroughly in a live environment, threw everything I could think of at it with PRT watching the server and rebooting.

Issue #181, opened 2024-06-27
Documentation for .envrc.dist


kristin-wiseman:
**Is your feature request related to a problem? Please describe.**
I'm trying to set up a local development environment for a Drupal 10 site hosted on Pantheon using Solr. I'm using DDEV with the [ ddev-drupal-solr add on](https://github.com/ddev/ddev-drupal-solr), and I've copied `.envrc.dist` into `.ddev/.env` . I've set the values according to the provided examples, but I'm still getting a warning message that the server is reachable but the core is not. I've tried setting `PANTHEON_INDEX_CORE` to both the core's name as well as its path, but neither work, and it's unclear what that variable expects. 

Also, there's no example for a `PANTHEON_INDEX_SCHEMA` variable, but I've found that setting one that matches the schema on the site on Pantheon removes the warning in the status report that says "This module only works when it is on a Pantheon environment or the environment variables are correctly set as described in .envrc.dist".

**Describe the solution you'd like**
Documentation within .envrc.dist that lists every variable necessary for the module to function in a local development environment. Also, comments for at least the `PANTHEON_INDEX_HOST` and `PANTHEON_INDEX_CORE` variables that describe the variable and demonstrate the expected format.

**Describe alternatives you've considered**
I've considered using a configuration split specifically for DDEV with a separate search server connected to the index. However, Search API Pantheon seems to prevent that server from working despite using the Standard Solr connector, and disabling it causes problems when switching between environments locally to export configuration.

stevector:
I was just on office hours with @kristin-wiseman trying to work though this

Setting environment variables of 

```
export PANTHEON_INDEX_HOST=${DDEV_HOSTNAME}
export PANTHEON_INDEX_PORT=8983
export PANTHEON_INDEX_PATH=/solr
export PANTHEON_INDEX_CORE="dev"
export PANTHEON_INDEX_SCHEME=http
```

We could get a message that Solr was reachable but not the core.
<img width="1237" alt="Screenshot 2024-07-17 at 3 14 11 PM" src="https://github.com/user-attachments/assets/5c59dcfb-5224-4ec8-8b57-10ac12b225e1">

Within the Pantheon Multidev we were checking against we saw that the Multidev had an absolute path for `PANTHEON_INDEX_CORE` (`/site/5f6a069b-e1e9-4c62-8606-d26c1038b080/environment/july2024`) but when we tried a similar absolute value from DDev (`/var/solr/data/dev`) we got the White Screen of Death
[...]

PR #182, opened 2024-07-19
WIP: Experimental hack for DDev

stevector:
I did some co-working today with @stovak and @jazzsequence trying to work out #181. We found that some requests from Drupal to Solr included the name of the core in the path of the request and some did not. (We did this by toggling `debug` property here: https://github.com/pantheon-systems/search_api_pantheon/blob/8.x/src/Services/PantheonGuzzle.php#L63)

Those that did not were 404s, those that did were 200s. So here's a hacky re-insertion of the core name that at least gets rid of the error in #181.
 
Screenshot 2024-07-18 at 9 48 59 PM

Someone with a better understanding of Solr than me could do more to explain. If I kept going, I'd stop comparing search_api_pantheon on a Pantheon multidev to search_api_pantheon on DDev and instead do a DDev to DDev comparison. I bet DDev works fine with regular search_api_solr.

### DIFF
--- a/src/Services/PantheonGuzzle.php
+++ b/src/Services/PantheonGuzzle.php
@@ -167,7 +167,18 @@ class PantheonGuzzle extends Client implements
         return !empty($item);
     });
     $uri = $uri->withPath('/' . ltrim(implode('/', $path_parts), '/'));
-    return $request->withUri($uri);
+
+    // This is an experimental hack to get Solr + DDev working.
+    // Some calls to Solr do not include the name of the core in the URL.
+    // I don't know why some do and some don't.
+    // But the errors seem to go away when the core name is included.
+    $path = $uri->getPath();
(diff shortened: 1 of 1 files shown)


kristin-wiseman:
@stevector Thank you so much for this! I copied your changes into my local version of the module, and my local Solr is no longer reporting errors and can even index content. I also wanted to mention that DDEV with both Solr add-on versions (Solr 8 [ddev-drupal-solr](https://github.com/ddev/ddev-drupal-solr) and cloud Solr [ddev-solr](https://github.com/ddev/ddev-solr)) works with regular search_api_solr.

stevector:
`[src/Services/PantheonGuzzle.php](https://github.com/pantheon-systems/search_api_pantheon/pull/182/changes#diff-6e0cc813562ab8022f9ea9be6bf3199d74490dab4571ada59e9602625492765f)` This file no longer exists

PR #189, opened 2024-11-08
[SITE-2457] Reload core during post schema.

kporras07:
- Reloads schema after schema is posted
- Adds reload terminus command
- calls new reload endpoint as part of `reloadCore()` instead of the no-op log.


src/Services/Endpoint.php
```diff
@@ -198,6 +198,22 @@ public function getSchemaUploadUri(): string {
...
   }
 
+  /**
+   * Get URL in pantheon environment to upload schema files.
+   *
+   * @return string
+   *   URL of envrionment.
+   */
+  public function getReloadUri(): string {
+    return vsprintf(
```
kporras07, line 208: this function should be reimplemented

src/Services/SchemaPoster.php
```diff
@@ -105,9 +106,55 @@ public function postSchema(string $server_id, $files = []): array {
...
     }
 
+    $status_code = $response->getStatusCode();
+    $this->logger->info('Status code: ' . $status_code);
+    if ($status_code >= 200 && $status_code < 300) {
+      // @TODO Maybe we need to capture exception here??
+      // Call reload on the server.
+      $this->reloadServer();
+    }
+    $this->logger->info('After server reload?');
```
kporras07, line 118: extra debugging logging we should remove

src/Services/Endpoint.php
stovak: Kevin is correct. You'll need to get the reload URL from the env var.

src/Services/SchemaPoster.php
stovak: I would leave ALL of the debugging in until you have a release ready. Also I could do a drush command to call the reload as well.

stovak (dismissed):
I think you need two things... 1. Change the way it gets the reload URL to the environment variable. 2. do a command to do the reload from drush.

stovak:
Also having a drush command will help you debug the process once we get to testing. Once you know the drush command works and you can trigger it at will, you can know that as long as the UI is triggering that function, all is working as it should.

kporras07:
I'm unsure whether reloading as a standalone command is needed/makes sense. AFAIK the goal is to make reload happen automatically on schema post

tests/Unit/EndpointServiceTest.php
```diff
@@ -43,6 +44,19 @@ public function testURIGeneration() {
       'one://two:1234/server-path/schema-path',
       $ep->getSchemaUploadUri()
     );
+    $this->assertEquals(
+      'one://two:1234/server-path/schema-path',
+      $ep->getSchemaUploadUri()
+    );
```
pwtyler:
```suggestion
```

src/Commands/Reload.php
```diff
@@ -0,0 +1,68 @@
...
+   * @command search-api-pantheon:reloadSchema
+   */
+  public function reloadSchema() {
+    try {
+      $this->schemaPoster->reloadServer();
+    }
+    catch (\Exception $e) {
+      $this->logger->error((string) $e);
+    }
+  }
```
pwtyler, line 66: This currently doesn't display anything, not sure the idiomatic way to write a drush command so this is bare bones.

.gitignore
```diff
@@ -8,3 +8,5 @@
 .idea
 .DS_Store
 docs/Badge.confluence
+
+.envrc
```
pwtyler:
```suggestion
.envrc

```

pwtyler (dismissed):
Been a number of hands on this. +1 to cut a release and get a fix out with the caveat that unit tests are failing (and were failing before this PR) and units are not running in CI.

pwtyler:
I've backed out the changes to PantheonSolrConnector (a6b1488 and its subsequent fixes), as we encountered an exception thrown from the dash running reload core at `/admin/config/search/search-api/server/pantheon_blog_solr8/solr-admin/reload-core`.

```
Message	RuntimeException: Unable to instantiate Schema Poster. in Drupal\search_api_pantheon\Plugin\SolrConnector\PantheonSolrConnector->reloadCore() (line 409 of /code/web/modules/contrib/search_api_pantheon/src/Plugin/SolrConnector/PantheonSolrConnector.php).
```

`drush search-api-pantheon:postSchema` and `/admin/config/search/search-api/server/pantheon_solr8/pantheon-admin/schema` seem fine though. still adovcating Reload Core may be best as follow-on work unless there's an important gap I'm missing here.

pwtyler (dismissed):

namespacebrian (dismissed):

pwtyler:
Replacing this with #192. `git diff c4383b0 65a5c94` shows no difference. That PR will be our workspace for future iteration on the currently broken "reload core".

PR #203, opened 2025-01-15
SITE-1240 - Add Drupal 11 compatibility

stovak:
Upgrade libraries and current code to work with drupal 11.

DEPRECATED by this release: 
* Drupal 9.x support
* Php in versions less than 8.1

- The target version tag for this release is 8.3 and the HEAD on 8.x.  

- Drupal 9.x installations should use version 8.2.x of the module until they upgrade to D10 or D11.

- Drupal 8.x was deprecated by a previous release


config/install/search_api.server.pantheon_solr8.yml
```diff
@@ -1,50 +0,0 @@
-langcode: en
```
kporras07, line 1: I'm worried about all of this config being removed and it's probably still referenced in the docs. Also: it's just "easier" for customers to have this as a starting point.

Is there any reason I'm missing on why you're removing this?

src/Commands/Diagnose.php
```diff
@@ -41,12 +38,10 @@ class Diagnose extends DrushCommands {
    *   Injected by container.
    */
   public function __construct(
-        LoggerChannelFactoryInterface $loggerChannelFactory,
         PantheonGuzzle $pantheonGuzzle,
         Endpoint $endpoint,
         SolariumClient $solariumClient
     ) {
-    $this->logger = $loggerChannelFactory->get('SearchAPIPantheon Drush');
```
kporras07, line 49: Is logger now there by default?

.github/workflows/ci.yml
```diff
@@ -45,15 +45,15 @@ jobs:
       - name: PHPCompatibility
         uses: pantheon-systems/phpcompatibility-action@v1
         with:
-          test-versions: 7.4-
+          test-versions: 8.2
   build_test:
     strategy:
       matrix:
-        drupal-version: [ 10 ]
+        drupal-version: [ 10, 11 ]
```
kporras07, line 56: If we claim D9 compatibility, I think we should also test there. I see some type changes that make me nervous about D9 :(

config/install/search_api.server.pantheon_solr8.yml
stovak: Agreed. See slack conversation.

rvtraveller:
Even if you make it past the test errors, it looks like there are still some type dependencies that need to be worked out. Seeing the below error on a site I'm working on upgrading to Drupal 11:

```TypeError: Drupal\search_api_pantheon\Services\PantheonGuzzle::sendRequest(): Argument #1 ($request) must be of type Psr\Http\Message\RequestInterface, Solarium\Core\Client\Request given, called in /code/web/modules/contrib/search_api_pantheon/src/Services/PantheonGuzzle.php on line 173 in Drupal\search_api_pantheon\Services\PantheonGuzzle->sendRequest() (line 98 of /code/web/modules/contrib/search_api_pantheon/src/Services/PantheonGuzzle.php)```

**Comment by stovak** [comment]
> Even if you make it past the test errors, it looks like there are still some type dependencies that need to be worked out. Seeing the below error on a site I'm working on upgrading to Drupal 11:
> 
> `TypeError: Drupal\search_api_pantheon\Services\PantheonGuzzle::sendRequest(): Argument #1 ($request) must be of type Psr\Http\Message\RequestInterface, Solarium\Core\Client\Request given, called in /code/web/modules/contrib/search_api_pantheon/src/Services/PantheonGuzzle.php on line 173 in Drupal\search_api_pantheon\Services\PantheonGuzzle->sendRequest() (line 98 of /code/web/modules/contrib/search_api_pantheon/src/Services/PantheonGuzzle.php)`

Sorry, this should have been a draft. It wasn't ready when you were checking it.

File: src/Commands/Diagnose.php (hunk 2 above)
Comment (stovak) [reply src/Commands/Diagnose.php:49]: Injected automatically.

**Review by namespacebrian — APPROVED** [review]

File: .github/workflows/ci.yml [hunk 4]
```diff
@@ -45,15 +45,15 @@ jobs:
...
       fail-fast: false
     runs-on: ubuntu-latest
     container:
-      image: quay.io/pantheon-public/build-tools-ci:8.x-php8.1
+      image: quay.io/pantheon-public/build-tools-ci:8.x-php8.2
```
pwtyler: Questions for follow on work: What are we using Build Tools for? Are we testing with a BT site, or just using the extra commands like workflow:wait (which is now in core).

Broadly leading to, can we switch to ubuntu-latest/shivammathur/setup-php/terminus-github-actions and drop the quay dep?

.codacy.yml
pwtyler, line 1: add to .gitattributes

phpcs.xml
pwtyler, line 1: add to .gitattributes ?

config/install/search_api.index.primary.yml
```diff
@@ -44,13 +39,21 @@ third_party_settings:
...
-description: 'An example index to get started with Solr 8 searching.'
+    multilingual:
+      limit_to_content_language: false
+      include_language_independent: true
+      use_language_undefined_as_fallback_language: false
+      specific_languages:
+        en: '0'
+      use_universal_collation: false
+id: primary
+name: primary
```
kporras07, line 55: Are there any references to index/server names in docs that we need to update now as a result of them changing names?

RoboFile.php
```diff
@@ -113,36 +176,35 @@ public function testFull(int $drupal_version = 9, string $site_name = NULL) {
 
     // Test all the Solr things.
     $this->testSolrEnabled($site_name);
-
-    // Test creating Solr index.
-    $this->testSolrIndexCreate($site_name, 'dev');
-
-    // Test running the reload
-    $this->testPantheonSolrReload($site_name, 'dev');
```
kporras07, line 121: Are you on purpose dropping these Solr and pantheon solr reload tests? I'm unsure what they entail but looks like some functionality we probably want to have tests on them unless I'm missing something

config/install/search_api.index.primary.yml
namespacebrian: the string `content_index` does not occur in the `documentation` repo code, if that helps at all..
nor does `Basic Content Index`..  `content.?index` case insensitive..

.github/workflows/ci.yml
stovak: should I do that here or in a separate PR?

RoboFile.php
namespacebrian: Looks like `testPantheonSolrReload ` has been removed entirely: https://github.com/pantheon-systems/search_api_pantheon/pull/203/files/a111b2fc51e923266d221373e014f1128e250012#diff-fed7dd361d1a8d696657705806d8e756ff49cc4659f2191ab0e91172a59be0d8L812

and  `testSolrIndexCreate` was mostly rewritten: https://github.com/pantheon-systems/search_api_pantheon/pull/203/files/a111b2fc51e923266d221373e014f1128e250012#diff-fed7dd361d1a8d696657705806d8e756ff49cc4659f2191ab0e91172a59be0d8L782-L839

.github/workflows/ci.yml
pwtyler: Separate PR is fine, here looking for confirmation on the lift.

composer.json
```diff
@@ -37,20 +37,21 @@
...
+        "php-http/guzzle7-adapter": "^1.1",
         "psr/event-dispatcher": "^1.0",
-        "symfony/finder": "^4.4|^5|^6",
-        "symfony/yaml": "^4.4|^5|^6"
+        "symfony/finder": "^4.4|^5|^6|^7",
+        "symfony/yaml": "^4.4|^5|^6|^7"
     },
     "require-dev": {
-        "consolidation/robo": "^3.0",
+        "consolidation/robo": "^4",
```
greg-1-anderson: This will constrain you to PHP 8.2. If you want to be able to test on 8.1 through 8.4, using `^3 || ^4` should work, as long as you are judicious about running `composer update` at the right time in CI

composer.json
```diff
@@ -15,7 +15,7 @@
         "preferred-install": "dist",
         "sort-packages": true,
         "platform": {
-            "php": "8.0"
+            "php": "8.2"
```
greg-1-anderson, line 18: Let's keep 8.1 for now.

pwtyler: Discussed the same in Slack as long as D10 and 8.1 are not EoL. 👍🏻

RoboFile.php
```diff
@@ -30,12 +33,77 @@ class RoboFile extends Tasks {
...
+  }
+
+  public function test() {
+    $this->output()->writeln('RoboFile constructor: ' . $this->started->format('Y-m-d H:i:s'));
+    $this->output()->writeln('Search index: ' . print_r($this->search_index, TRUE));
+    $this->output()->writeln('Search server: ' . print_r($this->search_server, TRUE));
+    $this->output()->writeln('Hello world!');
+  }
+
+  public function getInstallConfigs(): void {
```
greg-1-anderson, line 55: This code is very general-purpose, and yet at its root, it is still hard-coded to only support two configs. Seems like this would be much easier to read and maintain if it was just:
```
$this->search_index = Yaml::parseFile('./config/search_api.index.primary.yml');
$this->search_server = Yaml::parseFile('./config/search_api.server.pantheon_search.yml');
```
Wrapped with existing error handling. Even if you add more configs, you're better off adding one more line per variable than extending the abstraction.

composer.json
```diff
@@ -37,20 +37,21 @@
         }
     },
     "require": {
-        "php": ">=7.4",
+        "php": ">=8.2",
```
greg-1-anderson: Let's keep 8.1 for now.

greg-1-anderson (approved):
LGTM, with some small adjustments.

RoboFile.php
stovak: I will add them back after I'm sure they're needed. Right now, I just want to get the D11 release in the wild.

config/install/search_api.index.primary.yml
stovak: I'm loading the config installs at inception and referencing them by ID in the robofile

pwtyler (approved):

PR #204, opened 2025-01-16
Drupal 11 Compatibility

laura-johnson:
- Upgrade dependencies
- Do not override drush logger (causes fatal error)
- Use Symfony EventDispatcher instead of deprecated Drupal ContainerAwareEventDispatcher

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -46,8 +46,8 @@
         "http-interop/http-factory-guzzle": "^1.0",
         "kint-php/kint": "^4.1|^5",
         "psr/event-dispatcher": "^1.0",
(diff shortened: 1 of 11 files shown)


stovak:
Hey, Laura, thanks for this. Going to merge it into my branch where I'm doing essentially the same work. The CI scripts will only run on a local branch so I'll merge mine into the trunk.

stovak:
It may be a day or so before I get the CI working again... all of the CI was keyed to php 7.4 so I Gotta fix that before I merge. Thanks again.

laura-johnson:
Thanks for the update!

PR #218, opened 2025-06-30
[WIP]Add fix to migrate pantheon_solr8 search server to pantheon_search

rkunjappan:
When a user upgrade from Search API pantheon 8.2.x to 8.3.x  version 

- The update hook changes the search server ID from pantheon_solr8 to pantheon_search.
- All existing indexes currently configured with pantheon_solr8 as their search server will be migrated to pantheon_search.
- And then the old search  server with id 'pantheon_solr8' will be deleted.

Tested on Prod site.

### DIFF
--- a/search_api_pantheon.install
+++ b/search_api_pantheon.install
@@ -26,3 +26,39 @@ function search_api_pantheon_requirements($phase) {
 
   return $requirements;
 }
+
+/**
+ * Migrate  search server pantheon_solr8  to pantheon_search.
+ */
+function search_api_pantheon_update_8300() {
+  $config_factory = \Drupal::configFactory();
+  $server = $config_factory->getEditable('search_api.server.pantheon_solr8');
+  $logger = \Drupal::logger('search_api_pantheon');
+
+  if ($server->get('id')) {
+    $data = $server->getRawdata();
+    $data['id'] = 'pantheon_search';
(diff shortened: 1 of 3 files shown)


search_api_pantheon.install
```diff
@@ -26,3 +26,39 @@ function search_api_pantheon_requirements($phase) {
...
+  else {
+    return;
+  }
+
+  // Replace  server id  for all indexes with  pantheon_search.
+  $indexes = \Drupal::entityTypeManager()->getStorage('search_api_index')->loadMultiple();
+  foreach ($indexes as $index) {
+    if ($index->get('server') === 'pantheon_solr8') {
+      $index->set('server', 'pantheon_search');
+      $index->save();
```
dpagini, line 55: One problem that I think my team ran into with this is that changing the server like this... it causes all of our indexes to empty and need re-indexing. I think that will happen to teams here... not sure if it's this modules responsibility to re-index (probably not in an update hook), or if it's just worth calling it out on in release notes...? Not sure...

rkunjappan: @dpagini Thank you so much for the thoughtful feedback — it’s truly appreciated. This PR is still a work in progress, and we’ll be sure to clearly highlight the reindexing requirement in the release notes.

PR #221, opened 2025-07-07
[BUGS-9875]  #1 Add fix to upgrade from earlier versions to 8.3.x

rkunjappan:
**Issue**
The bug [BUGS-9875] is related to the changes we have done in 8.3.x version  of Search API pantheon module https://getpantheon.atlassian.net/browse/SITE-1240
[Drupal.org Issue](https://www.drupal.org/project/search_api_pantheon/issues/3505065)
In 8.3.0, the server id  of **Pantheon Search** was changed from **'pantheon_solr8'** to **'pantheon_search'**, and the  **Basic content index(content_index')**  in config/optional folder was replaced with a new index  **Primary (primary)** in config/install folder. 

**Fix added**

When a user upgrade from Search API pantheon 8.2.x to 8.3.x version,
- The update hook will update  the search server Id of Pantheon Search   from 'pantheon_solr8' to 'pantheon_search'.
- All existing indexes currently configured with 'pantheon_solr8' as their search server will be migrated to 'pantheon_search'.
- And then the old search server with id 'pantheon_solr8' will be deleted.

**_Reverted this changes_**
> 
> - Instead of updating the id of  the 'content_index' with 'primary' , a duplicate of 'content_index'  will be created and saved as 'Primary' index.
> 
> This is because , if 'content_index' has any associated view, search page , facets  or any other referenced entities, it will get broken. So by creating a duplicate of the 'content_index' , the new index will gets created, and if the customers want they can  move the associated entities for the old index to new primary index and remove the old content index manually later.
> 
> If the customers were not using the old index ('content_index') provided in the module,  they can still use the custom index they were using. 
> 
> NB: Whenever I tested the version 8.2.2, old content_index was never installed, even though I already had the dependencies( Search API, Search API Solr, Solr Backend) installed. So I tested the migration by manually creating an index with id 'content_index.

_**Attaching another PR where index id is replaced instead of creating the duplicate.**_

https://github.com/pantheon-systems/search_api_pantheon/pull/220

Created a site on prod and tested.

Screenshot 2025-07-10 at 5 58 28 PM

[BUGS-9875]: https://getpantheon.atlassian.net/browse/BUGS-9875?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/config/install/search_api.index.primary.yml
+++ b/config/install/search_api.index.primary.yml
@@ -51,9 +51,9 @@ third_party_settings:
       specific_languages:
         en: '0'
       use_universal_collation: false
-id: primary
+id: Primary
 name: primary
-description: ''
+description: 'Default Search Index for Pantheon Search'
 read_only: false
 field_settings: {  }
 datasource_settings:
--- a/config/install/search_api.index.primary.yml
+++ b/config/install/search_api.index.primary.yml
@@ -51,8 +51,8 @@ third_party_settings:
       specific_languages:
         en: '0'
       use_universal_collation: false
-id: Primary
-name: primary
+id: primary
+name: Primary
 description: 'Default Search Index for Pantheon Search'
 read_only: false
 field_settings: {  }
--- a/search_api_pantheon.install
+++ b/search_api_pantheon.install
@@ -45,7 +45,6 @@ function search_api_pantheon_update_8300() {
   $logger                 = \Drupal::logger('search_api_pantheon');
 
   // Skip migration if :
-  // - it's not an update from 8.2.x or earlier versions.
   // - 'pantheon_solr8' search server doesn't exists,
   // - or the 'pantheon_search' server already exists.
   if ($pantheon_solr8_config->isNew() || (!$pantheon_search_config->isNew())) {
--- a/search_api_pantheon.install
+++ b/search_api_pantheon.install
@@ -29,7 +29,6 @@ function search_api_pantheon_requirements($phase) {
   return $requirements;
 }
 
-
 /**
  * Implements hook_update_N().
  *
@@ -78,7 +77,7 @@ function search_api_pantheon_update_8300() {
       }
     }
     // Create a 'primary' index by duplicating 'content_index'.
-   //  This is executed only if 'content_index' (from prior versions) exists.
+    //  This is executed only if 'content_index' (from prior versions) exists.
     $old_index = Index::load($old_index_id);
     if ($old_index && $primary_index_config->isNew()) {
       $new_index = $old_index->createDuplicate();
--- a/search_api_pantheon.install
+++ b/search_api_pantheon.install
@@ -34,8 +34,7 @@ function search_api_pantheon_requirements($phase) {
  *
  * This update hook assists in the transition when upgrading the
  * module from earlier versions to 8.3.x.
- * It updates the search server from 'pantheon_solr8' to 'pantheon_search',
- * and creates a new 'primary' index if the 'content_index' (provided by earlier versions) exists.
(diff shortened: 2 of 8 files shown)


rkunjappan:
@greg-1-anderson @pwtyler Could you please review this?

search_api_pantheon.install
```diff
@@ -26,3 +28,84 @@ function search_api_pantheon_requirements($phase) {
...
+      $new_index->enforceIsNew();
+      $new_index->save();
+      $logger->info("Created new index 'Primary'. ");
+    }
+
+    // Commenting the code for deletion of old index,
+    // since the old index deletion should be done manually, if needed).
+    //
+    // $old_index->delete();
+    // $logger->notice('Old index @old deleted.', ['@old' => $old_index_id]);
```
rkunjappan: Do we need  this step, or Just notify users that no action is needed for the index change?"

config/install/search_api.index.primary.yml
```diff
@@ -51,9 +51,9 @@ third_party_settings:
       specific_languages:
         en: '0'
       use_universal_collation: false
-id: primary
+id: Primary
```
greg-1-anderson: What is the motivation for uppercasing the id? Wouldn't it be more typical to keep the id all lower, and uppercase the first character of the name?

rkunjappan: Oh, it was actually a mistake  I was trying to uppercase the first character of the name.

greg-1-anderson (approved):

search_api_pantheon.install
```diff
@@ -26,3 +27,64 @@ function search_api_pantheon_requirements($phase) {
...
+  $old_server_id          = 'pantheon_solr8';
+  $new_server_id          = 'pantheon_search';
+  $old_index_id           = 'content_index';
+  $config_factory         = \Drupal::configFactory();
+  $pantheon_solr8_config  = $config_factory->getEditable("search_api.server.pantheon_solr8");
+  $pantheon_search_config = $config_factory->getEditable("search_api.server.pantheon_search");
+  $logger                 = \Drupal::logger('search_api_pantheon');
+
+  // Skip migration if :
+  // - it's not an update from 8.2.x or earlier versions.
```
kporras07: Is this a condition actually being checked?

search_api_pantheon.install
```diff
@@ -26,3 +27,64 @@ function search_api_pantheon_requirements($phase) {
...
+  // - 'pantheon_solr8' search server doesn't exists,
+  // - or the 'pantheon_search' server already exists.
+  if ($pantheon_solr8_config->isNew() || (!$pantheon_search_config->isNew())) {
+    return;
+  }
+
+  try {
+    // Update the search server id 'pantheon_solr8' to  'pantheon_search' server
+    if ($pantheon_search_config->isNew()) {
```
kporras07, line 56: What makes this `new` in this context?

kporras07 (approved):
Left a couple of non blocking questions. LGTM :)

search_api_pantheon.install
rkunjappan: You're right.  The condition added is checking whether  the old server doesn't exist or new config exists which is supposed to be an 8.3.x install (That means not an update from earlier versions). I will remove the comment, if it's misleading.

search_api_pantheon.install
rkunjappan: The 'pantheon_search' is not an  already existing configuration.

PR #224, opened 2025-07-22
[BUGS-9875] #2 Fix for Pantheon Solr8 search server

rkunjappan:
Fix Added

- Added support for skipping the automatic migration from pantheon_solr8 to pantheon_search during database updates if the following line is set in settings.php:
`$settings['default_search_server'] = 'pantheon_solr8';`

- Updated the Solr endpoint logic to use the value defined in settings.php as the default search server. If no value is set, it will set default  name to 'pantheon_search'.

- Updated README command examples to allow passing a custom Solr server ID.
(If no Solr server ID is specified, pantheon_search will be used by default.)

@greg-1-anderson There are some codacy warnings, I will correct it, if the logic of the rest of the code looks okay to you.

### DIFF
--- a/search_api_pantheon.install
+++ b/search_api_pantheon.install
@@ -7,6 +7,7 @@
 
 use Drupal\search_api_pantheon\Plugin\SolrConnector\PantheonSolrConnector;
 use Drupal\search_api\Entity\Server;
+use Drupal\Core\Site\Settings;
 
 /**
  * Implements hook_requirements().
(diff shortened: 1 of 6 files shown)


search_api_pantheon.install
```diff
@@ -43,11 +44,13 @@ function search_api_pantheon_update_8300() {
...
   $pantheon_search_config = $config_factory->getEditable("search_api.server.pantheon_search");
   $logger                 = \Drupal::logger('search_api_pantheon');
+  $default_search_server  = Settings::get('default_search_server', NULL);
 
   // Skip migration if :
   // - 'pantheon_solr8' search server doesn't exists,
   // - or the 'pantheon_search' server already exists.
-  if ($pantheon_solr8_config->isNew() || (!$pantheon_search_config->isNew())) {
+  // - Default search server is set as 'pantheon_solr8' in settings.php
+  if ($pantheon_solr8_config->isNew() || (!$pantheon_search_config->isNew()) || $default_search_server === 'pantheon_solr8') {
```
kporras07, line 53: Should we really compare to the string or just check the value exists?

kporras07 (approved):
I love the idea of allowing the name to NOT change. Just left a small comment about one of the conditions in the if.

greg-1-anderson (approved):
This looks good to me. I don't think it's important to follow Codacy's advice regarding access to static methods. Reducing this from two warnings to one would be helpful, but it's not necessary to eliminate both, unless you find an easy way to do so.

search_api_pantheon.install
rkunjappan: @kporras07 Thank you for looking into this. I was a bit confused about whether I should compare the value to the specific string or just check if it exists. If the user sets a different value or makes a typo in settings.php for the default search server, and if the update hook  only checks for the existence of the value, it could unintentionally trigger a server migration, right?

rkunjappan:
> This looks good to me. I don't think it's important to follow Codacy's advice regarding access to static methods. Reducing this from two warnings to one would be helpful, but it's not necessary to eliminate both, unless you find an easy way to do so.

Thank you Greg

rkunjappan:
> I love the idea of allowing the name to NOT change. Just left a small comment about one of the conditions in the if.

Thank you Kevin

PR #213, opened 2025-05-05
[SITE-4690] Major cleanup

chx:
This removes overrides for Guzzle, Endpoint and the Solarium client as they are not needed. Pantheon specific endpoint functionality has been moved into the connector -- but it is still only 40% as long as the original. 

Robo based testing has been replaced with two shell scripts which do all the setup needed then generate five nodes and assert indeed five items are indexed. The get constraint functionality is a straight port from Robo, I did not throughly verify the logic or the git commands.

The search api server connector configuration fields are now visible just disabled. On local they are not disabled so it's much easier to develop with the module on locally now: just fill in local details and there you go. On Pantheon it overrides these settings.

The parameters and behavior for Drush commands were kept although I would recommend phasing out server_id in drush commands as it is not needed, the code now searches for the first server with pantheon connector to accommodate for the recent default server rename.

I did not touch the admin module materially, just did a coding style pass over it with phbcbf.

The Codacy issues are not new, in fact they are less severe than they used to be.

PR #230, opened 2025-09-15
[SITE-4690] Major cleanup

rkunjappan:
I created a new PR with the 8.4.x branch as the destination branch.
PR created by chx https://github.com/pantheon-systems/search_api_pantheon/pull/213

**Notes from CHX**

 This removes overrides for Guzzle, Endpoint and the Solarium client as they are not needed. Pantheon specific endpoint functionality has been moved into the connector -- but it is still only 40% as long as the original.

 Robo based testing has been replaced with two shell scripts, which do all the setup needed, then generate five nodes and assert indeed five items are indexed. The get constraint functionality is a straight port from Robo, I did not throughly verify the logic or the git commands.
 
 The Search API server connector configuration fields are now visible, but they’re disabled. Locally, however, they remain enabled, making development much easier: you can simply enter your local values and move on. On Pantheon, those settings are automatically overridden.
 
Screenshot 2025-12-02 at 2 32 02 PM

 The parameters and behaviour for Drush commands were kept although I would recommend phasing out server_id in drush commands as it is not needed, the code now searches for the first server with pantheon connector to accommodate for the recent default server rename.

**Changes done after Chx's major cleanup**

- Removed the Search API Pantheon Admin submodule, which depended on pantheon guzzle. Posting the schema is already handled by the 
`drush search-api-pantheon: postSchema` command.
https://github.com/pantheon-systems/search_api_pantheon/pull/213#issuecomment-2914454040

- Added several fixes from the 8.3.x branch including an update hook for server changes, with an option to skip search server migration. 

- To fix this issue https://www.drupal.org/project/search_api_solr/issues/3309224 , re-added the Stats Summary function in the Pantheon Solr Connector without using Guzzle. 

- Removed Kint usage from Drush commands.
 
- Fixed the Diagnose Drush command rewrite (https://getpantheon.atlassian.net/browse/SITE-5367)

- Updated Readme

 **Upgrade behavior**

Upgrading from 8.2.x / 8.3.x to 8.4.x is generally seamless.
But if customers are forced into the server migration path (using the update hook), they will need to reindex.


src/Commands/TestIndexAndQuery.php
```diff
@@ -121,31 +92,26 @@ public function testIndexAndQuery() {
...
       $this->logger->emergency("There's a problem somewhere...");
-      exit(1);
+      return self::EXIT_FAILURE;
     }
     catch (\Throwable $t) {
       \Kint::dump($t);
       $this->logger->emergency("There's a problem somewhere...");
-      exit(1);
+      return self::EXIT_FAILURE;
     }
```
pwtyler, line 127: Kint here and in `src/Commands/Diagnose.php` needs to be replaced now that it's no longer a dep.

.github/workflows/ci.sh
```diff
@@ -0,0 +1,35 @@
+#!/usr/bin/bash
+set -e
+. "$(dirname "${BASH_SOURCE[0]}")/git-constraint-helper"
+CONSTRAINT=$(get_current_constraint)
+terminus site:create "$SITE" "$SITE" "drupal-$DRUPAL_VERSION-composer-managed" --org "$TERMINUS_ORG"
```
pwtyler, line 5: (future work), this should be moved to creating a multidev of a named fixture site.

.github/workflows/ci.sh
```diff
@@ -0,0 +1,35 @@
...
+fi
+echo "Verifying the number of items in Solr directly"
+FOUND=$(terminus drush "$SITE.dev" -- search-api-pantheon:select '*' |grep -v notice|jq .response.numFound)
+if [ "$FOUND" != "5" ]; then
+  echo "Solr reports the number of documents is $FOUND, 5 was expected. See the following report for more info."
+  terminus drush "$SITE.dev" sapd
+  exit 1
+fi
+echo "Testing schema post"
+terminus drush "$SITE.dev" -- search-api-pantheon:postSchema /code/web/modules/contrib/search_api_solr/jump-start/solr8/config-set
```
pwtyler, line 35: Testing this succeeds (i.e. a 0 exit) is a good start, but there should be some assertion of the state after the fact (not a blocking request, but capture for future work if not addressed).

.github/workflows/ci.yml
```diff
@@ -95,22 +92,23 @@ jobs:
...
         run: |
-          export TERMINUS_ORG=$TERMINUS_ORG
-          ./vendor/bin/robo test:full $DRUPAL_VERSION $TERMINUS_SITE
+          ./.github/workflows/ci.sh
 
       - name: Cleanup sites
         if: ${{ always() }}
         run: |
-          ./vendor/bin/robo test:delete-sites
+          terminus -y site:delete $SITE
```
pwtyler, line 111: We should not delete always(), as that destroys environments we may want to review/investigate after a failed test. Changing this requires then having a proactive "ensure multidev exists" like what's included here https://github.com/pantheon-systems/terminus-github-actions/pull/31.

Can be held for future work.

src/GetPantheonSolrServerTrait.php
```diff
@@ -32,6 +32,7 @@ protected function getPantheonSolrServer(?string $server_id = NULL): ServerInter
     if (!$server_id) {
       $ids = $this->storage->getQuery()
         ->condition('backend_config.connector', 'pantheon')
+        ->condition('status', TRUE)
```
rkunjappan, line 35: Added a condition to return only the enabled Pantheon server. This prevents drush commands from selecting a disabled Pantheon server as the default server with the Pantheon connector.

src/Plugin/SolrConnector/PantheonSolrConnector.php
```diff
@@ -2,481 +2,234 @@
...
+  public function __construct(array $configuration, $plugin_id, array $plugin_definition) {
+    // While normally reading global state should be in ::create() the very
+    // point of this class is to force override the configuration with these
+    // environment variables. Testability is not a problem thanks to putenv()
+    // and more importantly, any testing would be pointless if it didn't
+    // assert these overrides.
+    if ($overrides = static::getEnvironmentVariables()) {
+      $configuration = $overrides + $configuration;
+      // This is used by ::createClient() only.
+      $configuration['search_api_pantheon_cert'] = ($_SERVER['HOME'] ?? '') . '/certs/binding.pem';
```
pwtyler: Should we be using the self-signed binding.pem at this point, or does mtlsproxy allow us to do away with this (and possibly PantheonCurl) entirely? I expect we should be using `pantheon_curl` in the prepend file if that means we can get away from even more of this.

src/Commands/Diagnose.php
```diff
@@ -21,207 +19,122 @@
...
+    if (empty($pantheon_yml['search']['version'])) {
+      // Merge from pantheon_upstream as fallback.
+      if (!empty($pantheon_upstream_yml['search']['version'])) {
+        $pantheon_yml['search']['version'] = $pantheon_upstream_yml['search']['version'];
+      }
+    }
+    if (empty($pantheon_yml['search']['version'])) {
+      // If still empty, throw an exception.
+      throw new \Exception('Unable to find search.version in pantheon.yml or pantheon.upstream.yml');
+    }
```
pwtyler, line 132: Today, the default solr version when solr is enabled with nothing in p.yml is solr 3. In the future, that will change, and we will enable solr 8 by default. Should we be concerned that this heathcheck will fail once that happens? Or leave that for a future version of search_api_pantheon?

src/Plugin/SolrConnector/PantheonSolrConnector.php
```diff
@@ -2,481 +2,234 @@
...
+    foreach (array_keys(static::getEnvironmentVariables()) as $key) {
+      if (isset($form[$key])) {
+        $form[$key]['#disabled'] = TRUE;
+        $form[$key]['#description'] = t('These fields are populated by Pantheon infrastructure".');
+      }
+    }
+    // Keep workarounds and advanced disabled on local too so when config is
+    // exported then there's no chance of it messing with Pantheon.
+    $form['workarounds']['#access'] = FALSE;
+    // @todo explore whether jts works.
```
pwtyler: If this todo is not captured anywhere, please open a spike.

search_api_pantheon.install
```diff
@@ -5,38 +5,17 @@
...
-
-  return $requirements;
-}
-
 /**
  * Implements hook_update_N().
  *
  * This update hook assists in the transition when upgrading the
- * module from earlier versions to 8.3.x.
+ * module from earlier versions to 4.x
```
pwtyler: This is now 8.4.x? or 8.3.x+?

src/Plugin/SolrConnector/PantheonSolrConnector.php
rkunjappan: Thank you @pwtyler. Created a spike ticket to track this https://getpantheon.atlassian.net/browse/SITE-5396

src/Plugin/SolrConnector/PantheonSolrConnector.php
rkunjappan: I tested removing both Pantheon binding.pem and PantheonCurl entirely, but that resulted in an SSL verification failure in GitHub CI tests for the self-signed certificate.
```
Drupal\search_api_solr\SearchApiSolrException while indexing item entity:node/5:en:
Solr endpoint https://mtlsproxyhost:9004/v1/ unreachable or returned unexpected
response code (code: 60, body:, message: Solr HTTP error: HTTP request failed,
SSL certificate problem: self-signed certificate (60)).
``` 
Unable to reproduce the error in already existing  Pantheon sites, but newly created Pantheon sites throw the same SSL error without the PantheonCurl class.


Removed pantheon binding.pem and kept PantheonCurl(after modification), which disables SSL verification for self-signed certificates. 
```
   public function createHandle($request, $endpoint): \CurlHandle {
    $handler = parent::createHandle($request, $endpoint);
    
    // Disable SSL peer verification for mtlsproxy's self-signed certificates.
    
    curl_setopt($handler, CURLOPT_SSL_VERIFYPEER, FALSE);
    
    return $handler;
  }


```
This fixes the above SSL verification error. 

In preprend file, I can see the SSL verification is disabled here https://github.com/pantheon-systems/cos-runtime-php/blob/main/php-prepend-includes/7.4-and-8-series/pantheon.php#L280

search_api_pantheon.install
rkunjappan: @pwtyler This is corrected now. Thank you for noticing it.

src/Solarium/PantheonSolrCurl.php
```diff
@@ -0,0 +1,53 @@
...
+  }
+
+  /**
+   * Get SSL options from Pantheon infrastructure.
+   *
+   * @return array
+   *   SSL verification options.
+   */
+  protected static function getPantheonCurlOptions(): array {
+    if (!function_exists('pantheon_curl_setup')) {
```
rkunjappan, line 34: Renamed the file name to PantheonSolrCurl.
Updated the Custom Curl  adapter code to use SSL verification options from  pantheon prepend file.

pwtyler (approved):

src/Commands/Diagnose.php
greg-1-anderson: I think we should allow empty solr version in pantheon.yml to pass. I didn't check to see if there were environment variables in the Appserver environment we could check to see the Solr version; if there were, that would be an even better solution.

src/Solarium/PantheonSolrCurl.php
greg-1-anderson: I kind of hate that we directly access `pantheon_curl_setup`, but I don't see a better option given the constraints today. Maybe as a follow-on task, you could add a comment in cos-runtime-php noting this external use of this function, so that future work doesn't remove it and break this module.

greg-1-anderson (approved):

src/Solarium/PantheonSolrCurl.php
pwtyler: this is the same thing we do for the drupal 7 module when running on Pantheon. What did you have in mind otherwise?

PR #231, opened 2025-09-26
[SITE-5266] perf: Move kint-php/kint to dev dependencies to eliminate production overhead

paulsheldrake:
## Summary

This PR moves  from runtime dependencies () to development dependencies () to eliminate significant performance overhead in production environments.

## Problem

Kint is currently loaded on every request in production, causing:
- **~27ms execution time** per request for composer metadata parsing
- **~3MB memory usage** for parsing 300+ package metadata 
- **Runs before Drupal caching** mechanisms can intercept

This overhead occurs because Kint's  file automatically parses  and  files to look for configuration, which happens during Composer's autoload process on every request.

## Solution

- **Move**  from  to  section
- **Remove** duplicate  entry from  section
- **Preserve** all debugging functionality in development environments
- **Eliminate** production overhead when using 

### Production (with )
✅ **Eliminates ~27ms + 3MB overhead per request**  
✅ **No functionality loss** (Kint shouldn't be in production)  
✅ **Faster page loads** and reduced memory usage

### Development (with dev dependencies)  
✅ **Kint remains fully available** for debugging  
✅ **No breaking changes** to development workflow  
✅ **All existing functionality preserved**

## Testing

- [x] Verified no runtime code uses Kint (only dev/test commands)
- [x] Confirmed RoboFile.php and Drush commands are dev-only tools
- [x] Tested that core module functionality is unaffected

## Performance Evidence

This change addresses a performance bottleneck identified via XDebug profiling where Kint's JSON parsing of composer metadata was consuming significant resources on every request before any application-level caching could help.

## Breaking Changes

None for typical usage. Only affects environments that:
1. Use  (production)  
2. AND try to use Kint debugging functions

Such usage would be incorrect anyway as debugging tools shouldn't be used in production.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -43,7 +43,6 @@
         "ext-zip": "*",
         "drupal/search_api_solr": "^4.3",
         "guzzlehttp/guzzle": "^7.5",
-        "kint-php/kint": "^5|^6",
         "php-http/guzzle7-adapter": "^1.1",
         "psr/event-dispatcher": "^1.0",
         "symfony/finder": "^4.4|^5|^6|^7",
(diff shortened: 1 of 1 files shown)


pwtyler:
Hi @paulsheldrake, thanks for the PR— We've actually got an overhaul (#230) of search_api_pantheon that includes removing kint as a dependency entirely. At the moment It's actually called in a couple of commands in (Diagnose.php and TestIndexAndQuery.php) even thought it really shouldn't, so just moving it to dev deps would break the current exception handling. Hoping to get that PR in and the major version published soon, it got sidelined for a bit but we're back on it.

PR #255, opened 2026-03-21
[SITE-5436] Add Solr9 support to  Search API Pantheon

rkunjappan:
## Summary

- Add Apache Solr 9 support alongside existing Solr 8 support
- Default server configuration updated to Solr 9 for **new installations**
- Solr version detection now uses `PANTHEON_SEARCH_VERSION` env var for pantheon environments
- Handle Solr 9 response format differences in `getStatsSummary()` with null-safe access
- Update Diagnose command to accept version 8 or 9
- Bump `drupal/search_api_solr` requirement from `^4.2` to `^4.3`
- Pin `drupal/coder` to `^8.3` to fix lint failures
- Expand CI matrix to test Solr 8 and 9 across Drupal 10 and 11

## Documentation

- New [UPGRADE.md](UPGRADE.md) with migration guides for 8.4.x → 8.5.x and Solr 8 → Solr 9, including rollback procedure

## Upgrade Path

- **Solr 8 → Solr 9 migration**: Switching versions provisions a new Solr core — **existing indexed data will not carry over**. Full reindex required. Update the module first, then change `pantheon.yml`. See [UPGRADE.md](UPGRADE.md) for step-by-step instructions.


src/Plugin/SolrConnector/PantheonSolrConnector.php
```diff
@@ -86,6 +87,46 @@ public static function getEnvironmentVariables(): array {
     ];
   }
 
+  /**
+   * Detects the Solr version from pantheon.yml or pantheon.upstream.yml.
+   *
+   * @return int
+   *   The Solr version (8 or 9). Defaults to 8 for backward compatibility.
+   */
+  protected static function detectSolrVersion(): int {
```
pwtyler: This value should not come from reading pantheon.yml on disk, but the env variable that processing pantheon.yml sets.

IIRC if that value is null/DNE it falls back to solr 3, we should confirm what we want the default behavior to be in conjunction with PRT since the default state is changing soon and we want to avoid a module update requirement.

UPGRADING.md
```diff
@@ -0,0 +1,199 @@
...
+5. **Clear the cache:**
+   ```bash
+   drush cr
+   ```
+
+6. **Test search functionality** thoroughly on non-production environments before deploying to live. Use `drush search-api-pantheon:diagnose` to verify the configuration.
+
+### Rolling Back to Solr 8
+
+To revert to Solr 8 after upgrading:
```
stevector:
```suggestion
If for some reason you need to revert to Solr 8 after upgrading:
```

UPGRADING.md
```diff
@@ -0,0 +1,199 @@
...
+
+1. Change `pantheon.yml` back to `version: 8`.
+2. Re-post the Solr 8 schema:
+   ```bash
+   drush search-api-pantheon:postSchema
```
stevector: we should add terminus to all the drush commands

rkunjappan: I will update the Drush commands. Thank you

rkunjappan (commented):
Remove this

src/Plugin/SolrConnector/PantheonSolrConnector.php
rkunjappan: Removed detectSolrVersion() and replaced it with the PANTHEON_SEARCH_VERSION env var, as suggested. The module no longer reads pantheon.yml directly — the platform processes it and sets the env var.                                            
                                                                                                                                                                                                                                                    
  **Local dev:**
                                                                                                                                                                                                                                                    
  The Pantheon connector only activates when PANTHEON_ENVIRONMENT is set, which only happens on Pantheon and Lando. Without Lando, developers configure Solr through the Drupal admin UI directly — no impact.                                      
                                                                                                                                                                                                                                                    
  For Lando users, Lando sets PANTHEON_ENVIRONMENT=lando  which activates the connector and locks the config form. Lando currently provides:                                                                                                         
  - PANTHEON_INDEX_HOST = "index" ✅
  - PANTHEON_INDEX_PORT = "449" ✅                                                                                                                                                                                                                  
 The following variables are set in Lando's code but missing from the docs. They may be set incorrectly — customers have reported Solr connection issues in the community Slack:                                                                                 
  - PANTHEON_INDEX_CORE                                                                                                                                                                                                                             
  - PANTHEON_INDEX_PATH                                                                                                                                                                                                                             
  - PANTHEON_INDEX_SCHEMA                                                                                                                                                                                                                           
  - PANTHEON_INDEX_RELOAD_PATH                                                                                                                                                                                                                      
  - PANTHEON_SEARCH_VERSION — (I added in fork PR:) https://github.com/lando/pantheon/compare/main...rkunjappan:pantheon:feat/update-search-env-variables
                                                                                                                                                                                                                                                    
  **Lando also has no Solr 9 support (only 3.6 and 8.8 images exist).**                                                                                                                                                                                 
                                                                                                                                                                                                                                                    
  Tracked in SITE-5661. 
  Need to work with the Lando maintainer to add Solr 9 and fix the missing variables.   
  
  (Sorry , for editing this comment many times)

src/Commands/Diagnose.php
```diff
@@ -131,8 +131,8 @@ public function verifyYamlFiles(): void {
       throw new \Exception('Unable to find search.version in pantheon.yml or pantheon.upstream.yml');
     }
 
-    if ($pantheon_yml['search']['version'] != '8') {
-      throw new \Exception('Unsupported search.version in pantheon.yml or pantheon.upstream.yml');
+    if (!in_array($pantheon_yml['search']['version'], ['8', '9'])) {
+      throw new \Exception('Unsupported search.version in pantheon.yml or pantheon.upstream.yml.');
```
pwtyler, line 135: Fine to update here and leave in place, but if we're to move towards managing with env variables rather than pantheon.yml, we should probably deprecate this command. For follow-on work.

pwtyler (approved):

AnaisPantheor (approved):

src/Commands/Diagnose.php
rkunjappan: @pwtyler, We have a follow-up ticket to improve the diagnose command: https://getpantheon.atlassian.net/browse/SITE-5446.

The command is useful for checking the overall health of Solr settings, especially when a site is using a misconfigured Solr config set or has related setup issues. 

I will improve the command to reflect the new changes.
Thank you

PR #261, opened 2026-04-24
SITE-5669: CI setup - Solr / search_api_pantheon tests 

ander-murane:
.github/workflows/multidev-tests.yml — multidev-based integration tests
- Trigger: PRs to 8.x + workflow_dispatch with optional Drupal/Solr version filters
- Matrix: Drupal 10/11 × Solr 8/9 = 4 parallel jobs, fail-fast: false
- Approach: Uses persistent test sites (from TEST_SITE_D10 / TEST_SITE_D11 secrets) with ephemeral multidevs created per run, instead of creating/destroying entire sites like the existing ci.yml
- Setup: Same container image, Terminus auth, and SSH pattern as the existing CI
- Execution: Delegates to run-multidev-tests.sh which handles multidev creation, field mapping tests (11 fields), environment parity tests
(8 assertions), and cleanup via EXIT trap


.github/workflows/multidev-tests.yml
```diff
@@ -0,0 +1,78 @@
+name: Multidev Tests
+on:
+  pull_request:
+    branches: [ 8.x ]
```
pwtyler:
```suggestion
    branches: [ 8.x, 8.5.x ]
```
beta on this branch

.github/workflows/multidev-tests.yml
```diff
@@ -0,0 +1,78 @@
...
+  multidev_test:
+    strategy:
+      matrix:
+        drupal-version: ${{ fromJSON(github.event.inputs.drupal_version && format('[{0}]', github.event.inputs.drupal_version) || '[10, 11]') }}
+        solr-version: ${{ fromJSON(github.event.inputs.solr_version && format('[{0}]', github.event.inputs.solr_version) || '[8, 9]') }}
+      fail-fast: false
+    runs-on: ubuntu-latest
+    container:
+      image: quay.io/pantheon-public/build-tools-ci:8.x-php8.1
+      options: --user root
```
pwtyler: Prefer ubuntu-latest and installing terminus/php with actions rather than qay build tools ci images.

scripts/manual-tests/test-field-mapping.sh
pwtyler, line 1: Are these no longer manual? Should these be moved to .github/scripts? Or what is the distinction?

ander-murane: The tests are now part of CI, unless you don't want them to be. I have deleted the scripts/manual-test directory

.github/scripts/test-field-mapping.sh
```diff
@@ -0,0 +1,218 @@
...
+  exit 1
+fi
+
+SITE_ENV="${TERMINUS_SITE}.${MULTIDEV_ENV}"
+FAILED=0
+
+# Clean up any existing test data from a previous run
+echo "::group::Cleanup existing test data"
+terminus drush "$SITE_ENV" -- ev "
+  \$nids = \Drupal::entityQuery('node')->condition('type', 'field_test')->accessCheck(FALSE)->execute();
```
stevector: @ander-murane passing PHP code as a string through drush is worrisome to me.

If every test run makes a new multidev, why do we need this cleanup at all? (if we do need to cleanup, we can put it elsewhere)

stevector (changes requested):
I'd like to review more but I'll first put in that the passing of PHP code through drush is an odd code smell. If we do need arbitrary PHP code for testing, we can put it in a test module.

.github/scripts/test-field-mapping.sh
ander-murane: Good point, removed.

ander-murane:
> I'd like to review more but I'll first put in that the passing of PHP code through drush is an odd code smell. If we do need arbitrary PHP code for testing, we can put it in a test module.

@stevector    Moved Drupal scaffolding (content type, 10 fields, test node) from
    per-multidev creation in test-field-mapping.sh to pre-configured on
    the dev environments. Multidevs inherit this via database clone.
    
    Tests passing: http://github.com/pantheon-systems/search_api_pantheon/actions/runs/26190313810/job/77060418654

pwtyler (changes requested):
Need to document and/or script re-creating the fixture environment (i.e. setting up the dev env) so that it could be recreated if needed. Link from readme (if its not in the readme already).

PR #262, opened 2026-04-30
[SITE-5198] PHP 8.5 compatibility and CI rework

mehta-asim:
## Summary

- PHP 8.5 compatibility fixes: Coder 9.0 sniff renames, deprecated `curl_close()` removal, `list()` modernization, orphaned annotation cleanup
- CI rework: replace full site creation (`ci.sh`, `git-constraint-helper`) with multidev-from-permanent-base-site pattern
- PHP version matrix (D10: 8.1/8.4, D11: 8.3/8.5) with Solr version matrix (8, 9)
- CI scripts: `create-multidev.sh`, `run-tests.sh`, `cleanup-multidevs.sh`, `set-multidev-and-constraint.sh`
- Cleanup: delete current multidev on success only, always sweep stale orphans (prefix match + 72h age-based)
- Show multidev URL on failure for investigation
- Updated actions: `actions/checkout` v4 -> v6.0.2, `webfactory/ssh-agent` v0.9.1 -> v0.10.0

## Context

[SITE-5198](https://getpantheon.atlassian.net/browse/SITE-5198)

PHP 8.5 compatibility for search_api_pantheon. Also reworks CI to use the multidev pattern (consistent with Content Publisher) instead of creating full sites per run. Uses permanent base sites with multidevs for each CI run.

## Prerequisites

- Base sites created with Solr enabled (actual Pantheon names may include random suffix)
- GitHub repo variables: `PANTHEON_SITE_D10`, `PANTHEON_SITE_D11` (must match actual Pantheon site names)
- GitHub repo secrets: `SANDBOX_SSH_KEY` (passphrase-free RSA), `GIT_EMAIL`, `TERMINUS_TOKEN`
- TERMINUS_TOKEN user must have access to both base sites

## Test plan

- [ ] `PANTHEON_SITE_D10`/`D11` repo vars match actual Pantheon site names
- [ ] All 8 matrix jobs pass (4 PHP combos x 2 Solr versions)
- [ ] "Show PHP versions" step confirms correct PHP on multidev
- [ ] Solr integration tests pass (schema post, index, query)
- [ ] On success: current multidev deleted, stale orphans cleaned
- [ ] On failure: multidev URL shown, env preserved for investigation
- [ ] Mirror job still works

[SITE-5198]: https://getpantheon.atlassian.net/browse/SITE-5198?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


.github/workflows/ci.yml
```diff
@@ -44,92 +45,115 @@ jobs:
...
-      BASH_ENV: ~/.bashrc
       DRUPAL_VERSION: ${{ matrix.drupal-version }}
+      GIT_EMAIL: ${{ secrets.GIT_EMAIL }}
+      GITHUB_RUN_NUMBER: ${{ github.run_number }}
+      PANTHEON_CI_SSH_KEY: ${{ secrets.SANDBOX_SSH_KEY }}
+      PHP_VERSION: ${{ matrix.php-version }}
       SOLR_VERSION: ${{ matrix.solr-version }}
+      TERMINUS_SITE: ${{ matrix.drupal-version == 10 && vars.PANTHEON_SITE_D10 || vars.PANTHEON_SITE_D11 }}
+      TERMINUS_TOKEN: ${{ secrets.TERMINUS_TOKEN }}
+      TERMINUS_VERSION: ${{ matrix.php-version >= '8.4' && '4.1.4' || '3.5.1' }}
```
pwtyler: I think Terminus 4 dropped < 8.2 so we only need T3 for < 8.2

.github/workflows/ci.yml
```diff
@@ -44,92 +45,115 @@ jobs:
...
 
-      - name: Login
+      - name: Install Terminus
+        run: |
+          mkdir -p $HOME/.terminus/plugins
+          curl -L https://github.com/pantheon-systems/terminus/releases/download/${TERMINUS_VERSION}/terminus.phar --output terminus
+          chmod +x terminus
+          sudo mv terminus /usr/local/bin/terminus
+          terminus --version
+
```
pwtyler, line 109: There's a terminus action, use that.

.github/workflows/ci.yml
```diff
@@ -44,92 +45,115 @@ jobs:
...
-          ./.github/workflows/ci.sh
+        run: bash .github/scripts/run-tests.sh
 
-      - name: Cleanup sites
-        if: ${{ always() }}
-        run: |
-          terminus -y site:delete $SITE
+      - name: Cleanup old CI multidevs
+        if: always()
+        run: bash .github/scripts/cleanup-multidevs.sh "$MULTIDEV_PREFIX" "${MULTIDEV_ENV:-}"
```
pwtyler: Envs should not be deleted if tests fail (so we can investigate). Cleanup steps could/should look for old out-of-date environments though to account for this.

.github/workflows/ci.yml
mehta-asim: Done

.github/workflows/ci.yml
mehta-asim: I am confused, do you mean that?
- if test succeeds we want to clean up the env
- but if it fails env should not be deleted for investigation
- and stale orphans should always be deleted

pwtyler: correct. if test passes, good to delete the multidev. if test fails, we want the env available for investigation. but we also want a mechanism to cleanup test envs from past builds that are no longer needed (this could be as simple as a timestamp check in cleanups checking for additional envs to clear out).

mehta-asim:
```yml
- name: Delete current multidev
        if: success()
        run: terminus multidev:delete "$TERMINUS_SITE.${MULTIDEV_ENV}" --delete-branch --yes || true

- name: Cleanup stale CI multidevs
        if: always()
        run: bash .github/scripts/cleanup-multidevs.sh "$MULTIDEV_PREFIX" "${MULTIDEV_ENV:-}"
```

- the existing script was deleting older multidevs and retained the current multidev for all cases
- i added a step earlier which will delete the current multidev if the tests pass, covering all cases
- i updated the script to add a 72h check for stale orphans, this will cover cases when we remove testing for d10 and add d12, to delete stale d10 multidevs

```sh
 # Timestamp check: delete old CI-pattern multidevs from any matrix combo
  if [[ "$env" =~ $CI_PATTERN ]]; then
    CREATED=$(echo "$MULTIDEVS" | jq -r --arg e "$env" '.[$e].created // 0' 2>/dev/null)
    if [[ "$CREATED" -gt 0 && "$CREATED" -lt "$CUTOFF" ]]; then
      echo "Deleting old CI multidev (age > ${MAX_AGE_HOURS}h): $env"
      terminus multidev:delete "$TERMINUS_SITE.$env" --delete-branch --yes || true
    fi
  fi
  ```

.github/workflows/ci.yml
mehta-asim: Done

pwtyler (approved):

rkunjappan (approved):

PR #263, opened 2026-05-22
[SITE-5661] Fix Lando Solr connection

AnaisPantheor:
## Summary

- Read `PANTHEON_INDEX_SCHEME` env var instead of hardcoding `https` in `PantheonSolrConnector`. Lando sets this to `http` but the module ignored it, causing SSL errors on every Solr request.
- Preserve URL context (`/solr/`) on Lando. The module unconditionally cleared it, which is correct on the platform (custom proxy) but breaks standard Solr paths on Lando.

Both changes are backward-compatible: on the real Pantheon platform, `PANTHEON_INDEX_SCHEME` is always `https` and `PANTHEON_ENVIRONMENT` is never `lando`.

## Context

Every version of `search_api_pantheon` from 8.4.0 onward is broken on Lando. The 8.4.0 rewrite auto-detects Pantheon environments (including Lando) and overrides Solr connection settings, but hardcodes HTTPS and clears the URL context unconditionally. Customers report in the [community Slack thread](https://pantheon-community.slack.com/archives/C0A76CTGB1D/p1771426185934479) that reverting to 8.3.4 fixes it.

## How to test

1. Start a Lando Pantheon site with Solr enabled (`search: version: 8` in `pantheon.yml`)
2. Install the module from this branch:
   ```bash
   composer require drupal/search_api_pantheon:dev-SITE-5661-lando-connector-fix
   ```
3. Enable modules and clear cache:
   ```bash
   lando drush en search_api search_api_solr search_api_pantheon -y
   lando drush cr
   ```
4. Run diagnose:
   ```bash
   lando drush search-api-pantheon:diagnose
   ```
5. Expect all checks green:
   - `Index SCHEME Value: http` (not `https`)
   - `Server Connection: The Solr server could be reached.`
   - `Core Connection: The Solr core could be accessed`
   - `Solr core URI: http://index:8983/solr/#/lando` (not a double-slash or missing `/solr/`)
6. Create content, index, and verify search returns results:
   ```bash
   lando drush eval "\$n = \Drupal\node\Entity\Node::create(['type'=>'page','title'=>'Test Search Content','body'=>[['value'=>'Solr integration testing on Lando','format'=>'basic_html']]]); \$n->save();"
   lando drush search-api:index
   lando drush eval "\$i = \Drupal\search_api\Entity\Index::load('primary'); \$q = \$i->query(); \$q->keys('Solr'); \$r = \$q->execute(); echo \$r->getResultCount() . ' results';"
   ```
   Expect: 1 or more results returned (not 0, not an error).

**Without this fix**, step 4 fails with `SSL routines::wrong version number`.

## Test results

| Solr | Module | Drupal | Result |
|---|---|---|---|
| 8 | 8.5.0-beta1 + this fix | D11 | PASS -- connectivity + search returns results |
| 9 | 8.5.0-beta1 + this fix | D11 | PASS -- same |
| 9 | 8.5.0-beta1 + this fix | D10 | PASS -- indexed 3 nodes, searched "WordPress migration" (1 result), "Solr" (2 results), "nonexistent" (0 results) |


AnaisPantheor:
## CI test failures

The test failures are not caused by this PR. All 6 failing "Run tests" jobs report `Search API reports 6 indexed documents, expected 5`.

The fixture sites (`search-api-pantheon-d10` and `search-api-pantheon-d11`) have a pre-existing "Field Mapping Test Node" on dev (created May 20 by the `search-ci` branch, commit `147e5c9`). Multidevs clone from dev, inherit that node, then `genc 5` produces 6 total items.

This is already fixed on the `search-ci` branch (`9c24611` -- `run-tests.sh` now counts before/after instead of hardcoding 5). CI will pass once that branch merges to `8.5.x`.

This connector PR only changes behavior when `PANTHEON_ENVIRONMENT === 'lando'`, which is never true on the platform.

rkunjappan:
Looks good to me! Good Job, Anais :-)

rkunjappan (approved):

mehta-asim (approved):

PR #276, opened 2026-07-14
fix: [BUGS-10912] prevent update hook from cascade-deleting search indexes

### LINKED ISSUE #246: Pantheon module update hook is deleting config entity.
**Describe the bug**
Hi team 
We have an config override service like below in a custom module which change modify the server in all indexes.
```
    if (!defined('PANTHEON_ENVIRONMENT')) {
      // Get any search_api index configurations.
      $indicies = array_filter($names, function ($name) {
        return str_starts_with($name, 'search_api.index.');
      });
      foreach ($indicies as $index) {
        $overrides[$index]['server'] = 'search_api_solr_drupal_server';
      }
}
```
We are using above config override to update the index config when the environment is non-pantheon(for eg. CI/CD pipeline). This will set the server id to `search_api_solr_drupal_server` in non-pantheon enviroment.

so now when `search_api_pantheon_update_10080()` update hook https://github.com/pantheon-systems/search_api_pantheon/blob/3b71962389a4575344441b7701f11ab1d17a9944/search_api_pantheon.install#L17  is deleting existing config entity in site.

Upon debugging I found https://github.com/pantheon-systems/search_api_pantheon/blob/ee5cefc9f408a282f4323dac2aa9a9249bc487c6/search_api_pantheon.install#L81 line on the update hook where the old server entity is getting deleted which is causing the existing index config object also deleted.

**To Reproduce**
1. Assume that you already have a site installed which has an index which use `pantheon_solr8` as server id
2. now update the search_api_pantheon module to 8.3.x or 8.4.x
3. now run the database update 
4. now if you try to get the index config through `drush cget search_api.index.[INDEX_ID]` it will not show anything

**Expected behavior**
Index should be available 

**Screenshots**

**Desktop (please complete the following information):**

**Smartphone (please complete the following information):**

**Additional context**

rkunjappan:
## Summary

Fix issue with Pantheon search server migration while using custom config override for search server  on non-Pantheon environments 

## Tested scenarios

| Scenario | Result |
|----------|--------|
| Site with `ConfigFactoryOverrideInterface` (Voya's setup on CI/local) | Migrated correctly, no cascade delete |
| Site without config override (normal Pantheon environments with pantheon search server) | Migrated correctly |
| Already-migrated site (re-run) | Skipped, idempotent |
| `$settings['default_search_server'] = 'pantheon_solr8'` | Skipped entirely |

Fixes #246

### DIFF
--- a/search_api_pantheon.install
+++ b/search_api_pantheon.install
@@ -5,7 +5,6 @@
  * Various hook implementations for installing the module.
  */
 
-use Drupal\search_api\Entity\Server;
 use Drupal\Core\Site\Settings;
 
 /**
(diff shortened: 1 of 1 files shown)


rkunjappan (commented):
```suggestion
  // Skip if old server doesn't exist (already migrated or using pantheon_search server),
  // or if the site explicitly keeps the old server name via settings.php.
  if (!$old_exists || $default_search_server === 'pantheon_solr8') {
    return;
  }
```

mehta-asim (approved):

tundeafolabi (approved):
lgtm

PR #283, opened 2026-08-25
[SITE-6015] #2 Add search_api_pantheon_cache submodule 

rkunjappan:
## Summary

- Adds `search_api_pantheon_cache` submodule that re-invalidates Solr-backed Views cache after `autoSoftCommit` fires (~5s), preventing stale search results
- Module enabled = feature active. No config toggle — zero overhead for sites that don't need it
- Alternative to the config-toggle approach in #282, better for cases where only specific sites (e.g. Content Publisher) need this fix

### How it works

1. When Drupal invalidates `search_api_list:*` cache tags (entity save), the submodule schedules a re-invalidation 6 seconds later via State API
2. On the next uncached request after the delay, `onRequest()` fires the re-invalidation, flushing any stale cached Views that were rendered before Solr committed

### Race condition 

Drupal invalidates cache tags at entity-save time, but Solr's `autoSoftCommit` doesn't make the new document visible for ~5 seconds. Any visitor who hits a Solr-backed View during that window gets stale results cached in Drupal's render cache.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -32,6 +32,18 @@
             "@code:lint"
         ]
     },
+    "autoload": {
+        "psr-4": {
+            "Drupal\\search_api_pantheon\\": "src/",
+            "Drupal\\search_api_pantheon_cache\\": "modules/search_api_pantheon_cache/src/"
+        }
+    },
+    "autoload-dev": {
+        "psr-4": {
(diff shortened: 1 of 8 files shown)


rkunjappan:
Going with the config-toggle approach in #282 instead — better discoverability, simpler maintenance.

